PR #4, opened 2018-04-12
Fix rollback error on roleset not being created properly. 

emilymye:


plugin/rollback.go
```diff
@@ -157,9 +157,9 @@ func (b *backend) serviceAccountPolicyRollback(ctx context.Context, req *logical
 		return err
 	}
 
-	// Take our any bindings still being used by this role set from roles being removed.
+	// Take out any bindings still being used by this role set from roles being removed.
 	rolesToRemove := util.ToSet(entry.Roles)
-	if rs.AccountId.ResourceName() == entry.AccountId.ResourceName() {
+	if rs != nil && rs.AccountId.ResourceName() == entry.AccountId.ResourceName() {
```
jefferai: if rs is nil is there any point in continuing on with the rest of the function? Or can we check a bit higher up and bail out early?

emilymye: If it's nil, we should actually clean up - this part checks that we don't accidently delete bindings for a roleset that still exist (i.e. someone deleted and re-added bindings but the WAL entry was not cleared yet).

jefferai: Ah, gotcha. Looks good then!

emilymye:
#3

PR #10, opened 2018-04-26
Fix TTL when config has not been set. 

emilymye:
Was panic on nil - adding reviewers for now since I'm not sure what the TTL changes were.

### DIFF
--- a/plugin/secrets_service_account_key.go
+++ b/plugin/secrets_service_account_key.go
@@ -170,6 +170,9 @@ func (b *backend) getSecretKey(ctx context.Context, s logical.Storage, rs *RoleS
 	if err != nil {
 		return nil, errwrap.Wrapf("could not read backend config: {{err}}", err)
 	}
+	if cfg == nil {
+		cfg = &config{}
+	}
 
 	iamC, err := newIamAdmin(ctx, s)
 	if err != nil {


chrishoffman:
This is not required as passing in the zero values will default to these values in core. Are you seeing an issue with incorrect values being set?

plugin/secrets_service_account_key.go
```diff
@@ -99,7 +99,10 @@ func (b *backend) secretKeyRenew(ctx context.Context, req *logical.Request, d *f
 		return nil, err
 	}
 	if cfg == nil {
-		cfg = &config{}
+		cfg = &config{
```
chrishoffman: This change is not required.  You can just pass the zero values back and they will get set appropriately.

plugin/secrets_service_account_key.go
```diff
@@ -170,6 +173,12 @@ func (b *backend) getSecretKey(ctx context.Context, s logical.Storage, rs *RoleS
 	if err != nil {
 		return nil, errwrap.Wrapf("could not read backend config: {{err}}", err)
 	}
+	if cfg == nil {
+		cfg = &config{
```
chrishoffman: You can just initialize an empty struct here.  No need to query the system values.

chrishoffman: I missed this panic on first pass and that was the reason for my initial comment.

jefferai:
@emilymye The TTL changes were a part of us going through all backends and plugins and modifying them so that Vault's core, not each backend, is now in charge of actual TTL issue/renewal calculations. This is safer and more standardized but required some updating of code in most backends.

plugin/secrets_service_account_key.go
jefferai: But, you still don't need to set the values :-)

emilymye:
SG, removed the values - I'm going to go ahead and merge!

jefferai:
👍

PR #11, opened 2018-04-26
Switch IAM to use the discovery doc API

emilymye:
Better handling of nonsense from dealing with ALL THE APIS:

- Previous implementation was using a technically deprecated field 'flatPath' (most of the docs still use it?) so we stopped using it
- Storage does whatever they want re: REST URL format and request format
- The discovery doc for compute straight up just has the wrong setIamPolicy request format.
 
Also:
- More tests
- Some renaming


emilymye:
I added more tests and confirmed this worked with a good spread of APIs' resources: 
- cloudkms
- cloudresourcemanager
- iam
- spanner
- storage
- pubsub
 
Need to get access in my test account to check the rest but the most problematic ones are in this list so I'm ok with moving forward with these verified for now.

PR #15, opened 2018-07-30
Make sure revoke works if token has already been revoked internally

### LINKED ISSUE #14: Not able to remove expired lease for Access Token
Hi, 

I have seen many stale leases for GCP access token, and I am not able to clean them up:
```
[ERROR] secrets.system.system_6aeff124: lease revocation failed: lease_id=gcp/token/tsk8s-dde/3614e5d1-6cc2-c7cb-75eb-bd86127facd7 error="failed to revoke entry: resp: &logical.Response{Secret:<nil>, Auth:<nil>, Data:map[string]interface {}{"error":"googleapi: got HTTP response code 400 with body: {\n  \"error\": \"invalid_token\"\n}"}, Redirect:"", Warnings:[]string(nil), WrapInfo:(*wrapping.ResponseWrapInfo)(nil)} err: <nil>"
```

By the way, should these leases be removed automatically since they are not refreshable?

Thanks

emilymye:
Check for error returned by oauth server if token has already been revoked by Google.

Fixes #14. 

The revoke request only returns status codes 200/400, hence error body checking.

### DIFF
--- a/plugin/secrets_access_token.go
+++ b/plugin/secrets_access_token.go
@@ -16,6 +16,8 @@ import (
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/googleapi"
 	"google.golang.org/api/iam/v1"
+	"strings"
+	"log"
 )
 
 const (
@@ -80,6 +82,11 @@ func (b *backend) secretAccessTokenRenew(ctx context.Context, req *logical.Reque
 	return logical.ErrorResponse("short-term access tokens cannot be renewed - request new access token instead"), nil
 }
 
+func isInvalidTokenErr(err error) bool {
+	gErr, ok := err.(*googleapi.Error)
+	return ok && gErr.Code == 400 && strings.Contains(gErr.Body, "invalid_token")
+}
+
 func secretAccessTokenRevoke(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
 	tokenRaw, ok := req.Secret.InternalData["access_token"]
 	if !ok {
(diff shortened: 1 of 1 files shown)


jefferai:
I'm not sure this is a good idea. It essentially swallows an error of the user's, which is that they are managing Vault-managed user accounts outside of Vault.

plugin/secrets_access_token.go
```diff
@@ -91,6 +93,17 @@ func secretAccessTokenRevoke(ctx context.Context, req *logical.Request, d *frame
...
 	}
 	if err := googleapi.CheckResponse(resp); err != nil {
+		gErr, ok := err.(*googleapi.Error)
+		if !ok {
+			return logical.ErrorResponse(err.Error()), nil
+		}
+		if gErr.Code == 400 && strings.Contains(gErr.Body, "invalid_token") {
+			// The token is no longer valid, probably because it was revoked internally
+			// Mark as revoked but return a warning.
+			return &logical.Response{
```
sethvargo: It's going to make the code more complicated, but I think we should add another field to the schema that's like `ignore_failed_revocation` that is false by default. 

If that field is set, print the warning.

If that field is not set, return an error with a suggestion to set that field on the request.

emilymye: sgtm

emilymye: On second thought, I think this is something that should be expected - it's likely Vault will always attempt to revoke the lease and will get this error as long as the user doesn't revoke it during its lifetime. Updated to just check that this doesn't happen and ignore it.

plugin/secrets_access_token.go
```diff
@@ -80,18 +82,29 @@ func (b *backend) secretAccessTokenRenew(ctx context.Context, req *logical.Reque
 	return logical.ErrorResponse("short-term access tokens cannot be renewed - request new access token instead"), nil
 }
 
+func isInvalidTokenErr(err error) bool {
+	gErr, ok := err.(*googleapi.Error)
+	return ok && gErr.Code == 400 && strings.Contains(gErr.Body, "invalid_token")
```
jefferai: Quick check -- this would be a 400 not a 403? 403 is likely more proper but that doesn't mean it's what Google's using :-)

plugin/secrets_access_token.go
```diff
@@ -80,18 +82,29 @@ func (b *backend) secretAccessTokenRenew(ctx context.Context, req *logical.Reque
...
+	if err == nil {
+		err = googleapi.CheckResponse(resp)
 	}
-	if err := googleapi.CheckResponse(resp); err != nil {
-		return logical.ErrorResponse(err.Error()), nil
+
+	if err != nil {
+		// Token may have already expired on server; ignore if OAuth server returns error.
+		if req.Secret.ExpirationTime().Before(time.Now()) && isInvalidTokenErr(err) {
+			log.Printf("[WARNING] Manual token revocation failed - this is usually because token has already been invalidated. Ignored error: %v", err)
```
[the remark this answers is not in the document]
sethvargo: I thought we'd be warning this to the caller too, as part of the response **in addition to** logging it in the server logs.

plugin/secrets_access_token.go
emilymye: Yup! This server will only return 400 or 200 and nothing else! So great!!

PR #20, opened 2018-09-17
Add possible quick fix for #14

### LINKED ISSUE #14: Not able to remove expired lease for Access Token

emilymye:
Fixes #14. 

Temporary change, access_token behavior is going to change soon.

### DIFF
--- a/plugin/secrets_access_token.go
+++ b/plugin/secrets_access_token.go
@@ -23,7 +23,7 @@ import (
 const (
 	SecretTypeAccessToken     = "access_token"
 	revokeAccessTokenEndpoint = "https://accounts.google.com/o/oauth2/revoke"
-	revokeTokenWarning        = `revocation request was successful; however, due to how OAuth access propagation works, the OAuth token might still be valid until it expires`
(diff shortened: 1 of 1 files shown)


emilymye:
@jefferai @briankassouf fyi - this will probably get overridden by previously discussed changes. In any case we should probably just remove the time check for invalid tokens since they have been "revoked".

PR #21, opened 2018-09-18
Remove token leasing/revocation

emilymye:
* Remove token leasing and revocation (i.e. don't return logical.Secret anymore)
* Also switched order of roleset deletion because IAM's deletion behavior for nonexistant keys switched from returning 404 to 403 (tests were failing)

TODO: Extensive documentation and warnings/notices for users potentially impacted. 

tl;dr: OAuth service account tokens do not allow for revocation (revocation requests silent no-op but return 200s)

### DIFF
--- a/plugin/secrets_access_token.go
+++ b/plugin/secrets_access_token.go
@@ -75,7 +75,7 @@ func (b *backend) secretAccessTokenResponse(ctx context.Context, s logical.Stora
 	return &logical.Response{
 		Data: map[string]interface{}{
 			"token":              token.AccessToken,
-			"token_ttl":          time.Now().UTC().Sub(token.Expiry.UTC()),
+			"token_ttl":          token.Expiry.UTC().Sub(time.Now().UTC()),
 			"expires_at_seconds": token.Expiry.Unix(),
 		},
 	}, nil
--- a/plugin/secrets_access_token.go
+++ b/plugin/secrets_access_token.go
@@ -11,6 +11,7 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/iam/v1"
+	"time"
 )
 
 func pathSecretAccessToken(b *backend) *framework.Path {
@@ -73,8 +74,9 @@ func (b *backend) secretAccessTokenResponse(ctx context.Context, s logical.Stora
 
 	return &logical.Response{
 		Data: map[string]interface{}{
-			"token":      token.AccessToken,
-			"expires_at": token.Expiry,
+			"token":           token.AccessToken,
+			"token_ttl":       time.Now().Sub(token.Expiry),
+			"expires_at_unix": token.Expiry.Unix(),
 		},
 	}, nil
 }
@@ -106,10 +108,9 @@ func (tg *TokenGenerator) getAccessToken(ctx context.Context, iamAdmin *iam.Serv
 }
 
 const deprecationWarning = `
-We have recently changed the behavior of the gcp/token/ endpoint to no longer generate leases. 
-This is due to limitations of the GCP API, as service account OAuth access tokens cannot be revoked.
-This access_token secret and associated lease were created by an older version of the GCP secrets engine
-and will be cleaned up now. Note that there is the chance that this access_token, if not already expired, 
+This endpoint no longer generates leases due to limitations of the GCP API, as OAuth2 tokens belonging to Service 
+Accounts cannot be revoked. This access_token and lease were created by a previous version of the GCP secrets 
+engine and will be cleaned up now. Note that there is the chance that this access_token, if not already expired, 
 will still be valid up to one hour. 
 `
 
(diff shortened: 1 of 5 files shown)


emilymye:
cc @sethvargo

plugin/secrets_access_token.go
```diff
@@ -141,20 +71,12 @@ func (b *backend) getSecretAccessToken(ctx context.Context, s logical.Storage, r
...
-	}
-	resp := b.Secret(SecretTypeAccessToken).Response(secretD, internalD)
-	resp.Secret.TTL = token.Expiry.Sub(time.Now())
-	resp.Secret.Renewable = false
-
-	return resp, err
+	return &logical.Response{
+		Data: map[string]interface{}{
+			"token":      token.AccessToken,
+			"expires_at": token.Expiry,
```
sethvargo: What type is this? In my experience, date types tend not to be transported well.

Additionally, if we calculate this as a TTL (time.Now().Unix().Sub(token.Expiry) and return this as a field named `*_ttl`, the Vault CLI will format it nicely (i.e. 30m)

plugin/secrets_access_token.go
```diff
@@ -182,3 +104,63 @@ func (tg *TokenGenerator) getAccessToken(ctx context.Context, iamAdmin *iam.Serv
 	}
 	return tkn, err
 }
+
+const deprecationWarning = `
+We have recently changed the behavior of the gcp/token/ endpoint to no longer generate leases.
```
sethvargo: Suggested:

```text
This endpoint no longer generates leases due to limitations of the GCP API, as OAuth2 tokens belonging to Service 
Accounts cannot be revoked. This access_token and lease were created by a previous version ...
```

plugin/secrets_access_token.go
```diff
@@ -182,3 +104,63 @@ func (tg *TokenGenerator) getAccessToken(ctx context.Context, iamAdmin *iam.Serv
...
+This path will generate a new OAuth2 access token for accessing GCP APIs.
+A role set, binding IAM roles to specific GCP resources, will be specified 
+by name - for example, if this backend is mounted at "gcp",
+then "gcp/token/deploy" would generate tokens for the "deploy" role set.
+
+On the backend, each roleset is associated with a service account. 
+The token will be associated with this service account. Tokens have a 
+short-term lease (1-hour) associated with them but cannot be renewed.
+
+NOTE: As of Vault 0.11.2, we have changed the behavior of the token/ endpoint.
```
sethvargo: I don't think we should have this in help text. The warning suffices, and I don't think we should be carrying history like this.

plugin/secrets_access_token.go
```diff
@@ -182,3 +104,63 @@ func (tg *TokenGenerator) getAccessToken(ctx context.Context, iamAdmin *iam.Serv
...
+	resp := logical.ErrorResponse("short-term access tokens cannot be renewed - request new access token instead")
+	resp.AddWarning(deprecationWarning)
+	return resp, nil
+}
+
+// Revoke will no-op and pass but warn the user. This is mostly to clean up old leases.
+// Any associated secret (access_token) has already expired and thus doesn't need to
+// actually be revoked,  or will expire within an hour and currently can't actually be revoked anyways.
+func (b *backend) secretAccessTokenRevoke(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
+	return &logical.Response{Warnings: []string{deprecationWarning}}, nil
```
sethvargo: Nit: I think we should always use the `AddWarning` API, since the field structs aren't technically part of the "API".

```go
var resp logical.Response
resp.AddWarning(deprecationWarning)
return &resp
```

plugin/secrets_access_token.go
emilymye: I left the expiry but changed it to a unix timestamp, and added token_ttl

plugin/secrets_access_token.go
emilymye: done

plugin/secrets_access_token.go
```diff
@@ -73,8 +74,9 @@ func (b *backend) secretAccessTokenResponse(ctx context.Context, s logical.Stora
 
 	return &logical.Response{
 		Data: map[string]interface{}{
-			"token":      token.AccessToken,
-			"expires_at": token.Expiry,
+			"token":           token.AccessToken,
+			"token_ttl":       time.Now().Sub(token.Expiry),
```
sethvargo: Thanky. In case you're curious, here's the magic: https://github.com/hashicorp/vault/blob/d3017e259fc23897896e2ab8c59cda8c9bdc0ae3/command/format.go#L249-L255

Do we need to make both of those times UTC for safety? `time.Now().UTC().Sub(token.Expiry.UTC())`?

plugin/secrets_access_token.go
```diff
@@ -73,8 +74,9 @@ func (b *backend) secretAccessTokenResponse(ctx context.Context, s logical.Stora
...
+			"expires_at_unix": token.Expiry.Unix(),
```
sethvargo: "Unix" could be interpreted as unix seconds or unix with microseconds or unix with milliseconds. I would recommend we just say "expires_at_seconds" or "expires_in_seconds"

PR #28, opened 2019-02-07
Roleset return data consistency 

broamski:
When reading a roleset, this changes one of the returned keys from 'service_account_project', to 'project' in order to match up with the input schema. This alignment is useful when using config management
tools to manage vault's configuration.


jefferai:
@emilymye Proably should evaluate whether the correct behavior is to fix input (with backwards compat) or to fix output.

emilymye:
I think this is fine - it probably should just be project.

jefferai:
@tyrannosaurus-becks ^ If we make this change here can we update the tf provider concurrently, if it's expecting the non-same name?

tyrannosaurus-becks:
Thanks @jefferai ! Looks like it's not a currently implemented endpoint in the Terraform Vault Provider so there's nothing to do on that end.

PR #30, opened 2019-03-07
Fix binding parsing (JSON-specific)

### LINKED ISSUE #27: Unable to create IAM role bindings for roleset using JSON formatting
This issue describes inconsistencies between the `bindings` property of creating a new roleset for the GCP Secrets backend when using HCL, versus when trying to use JSON. Take the following, confirmed working, HCL config for a roleset:
```
vault write gcp/roleset/test-role \
    project="${GOOGLE_CLOUD_PROJECT}" \
    bindings=@binding.hcl
```

Where `binding.hcl` is:

```
resource "//cloudresourcemanager.googleapis.com/projects/"${GOOGLE_CLOUD_PROJECT}" {
  roles = ["roles/viewer"]
}
```

When reading the config back in JSON format we receive:

```
vault read gcp/roleset/test-role -format=json
{
  "request_id": "<ID>",
  "lease_id": "",
  "lease_duration": 0,
  "renewable": false,
  "data": {
    "bindings": {
      "//cloudresourcemanager.googleapis.com/projects/<GOOGLE_CLOUD_PROJECT>": [
        "roles/viewer"
      ]
    },
    "secret_type": "service_account_key",
    "service_account_email": "vault<id-info>@<GOOGLE_CLOUD_PROJECT>.iam.gserviceaccount.com",
    "service_account_project": "<GOOGLE_CLOUD_PROJECT>"
  },
  "warnings": null
}
```

Now we use the above json to create another roleset, call it `test-role2`:
```
vault write gcp/roleset/test-role2 @test-role2.json
Success! Data written to: gcp/roleset/test-role2
```

Where the json in test-role2.json looks like:
``` 
{
    "project": "${GOOGLE_CLOUD_PROJECT},
    "secret_type": "service_account_key",
     "data": {
      "bindings": {
        "//cloudresourcemanager.googleapis.com/projects/${GOOGLE_CLOUD_PROJECT}": {
          "roles": [
            "roles/viewer"
          ]
        }
      }
    }
}
```

What we see is that while the roleset and the binding resource does get created, the list of IAM Roles is left blank:

```
vault read gcp/roleset/test-role2 -format=json
{
  "request_id": "<ID>",
  "lease_id": "",
  "lease_duration": 0,
  "renewable": false,
  "data": {
    "bindings": {
      "//cloudresourcemanager.googleapis.com/projects/<GOOGLE_CLOUD_PROJECT>": []
    },
    "secret_type": "service_account_key",
    "service_account_email": "vault<ID-info>@<GOOGLE_CLOUD_PROJECT>.iam.gserviceaccount.com",
    "service_account_project": "<GOOGLE_CLOUD_PROJECT>"
  },
  "warnings": null
}
```

Multiple permutations of the JSON were tried: both using JSON for the entire roleset config, and also only when using JSON for the bindings value itself. 

I suspect it has to do with how paths are being parsed - both line 351 and 358 [here](https://github.com/hashicorp/vault-plugin-secrets-gcp/blob/master/plugin/path_role_set.go#L351), as well as the way it is parsing the `roles` KV, [here](https://github.com/hashicorp/vault-plugin-secrets-gcp/blob/master/plugin/util/parse_bindings.go#L93), but any RCA or confirmation would be appreciated.

emilymye:
Use Value() which does quote handling. Fixes #27.

### DIFF
--- a/plugin/util/parse_bindings.go
+++ b/plugin/util/parse_bindings.go
@@ -88,12 +88,15 @@ func parseBindingObjList(topList *ast.ObjectList) (map[string]StringSet, error) 
 
 		resourceList := item.Val.(*ast.ObjectType).List
 		for _, rolesItem := range resourceList.Items {
-			key := rolesItem.Keys[0].Token.Value().(string)
-			switch key {
+			if len(rolesItem.Keys) == 0 || rolesItem.Keys[0].Token.Value() == nil {
+				merr = multierror.Append(merr, fmt.Errorf("invalid nil item under key %q in resource %q (line %d)", key, resourceName, rolesItem.Assign.Line))
+			}
+			roleKey := rolesItem.Keys[0].Token.Value().(string)
+			switch roleKey {
 			case "roles":
 				parseRoles(rolesItem, bindings[resourceName], merr)
 			default:
-				merr = multierror.Append(merr, fmt.Errorf("invalid key '%s' in resource '%s' (line %d)", key, resourceName, item.Assign.Line))
+				merr = multierror.Append(merr, fmt.Errorf("invalid key %q in resource %q (line %d)", roleKey, resourceName, rolesItem.Assign.Line))
 				continue
 			}
 		}
--- a/plugin/util/parse_bindings.go
+++ b/plugin/util/parse_bindings.go
@@ -152,6 +152,10 @@ func parseRolesObject(rolesObj *ast.ObjectItem, parsedRoles StringSet) error {
 }
 
 func parseRole(parent *ast.ObjectItem, roleNode ast.Node) (string, error) {
+	if roleNode == nil {
+		return "", fmt.Errorf(`unexpected empty role item (line %d)`, parent.Assign.Line)
+	}
+
 	roleLitType, ok := roleNode.(*ast.LiteralType)
 	if !ok || roleLitType == nil {
 		return "", fmt.Errorf(`unexpected nil item in roles list (line %d)`, parent.Assign.Line)
--- a/plugin/util/parse_bindings.go
+++ b/plugin/util/parse_bindings.go
@@ -74,7 +74,12 @@ func parseBindingObjList(topList *ast.ObjectList) (map[string]StringSet, error) 
 			continue
 		}
 
-		key := item.Keys[0].Token.Value().(string)
+		keyRaw := item.Keys[0].Token.Value()
+		key, ok := keyRaw.(string)
+		if !ok {
+			merr = multierror.Append(merr, fmt.Errorf("invalid resource item %v is not a string on line %d", keyRaw, item.Assign.Line))
+			continue
+		}
 		if key != "resource" {
 			merr = multierror.Append(merr, fmt.Errorf("invalid key '%s' (line %d)", key, item.Assign.Line))
 			continue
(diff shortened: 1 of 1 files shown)


plugin/util/parse_bindings.go
```diff
@@ -88,7 +88,7 @@ func parseBindingObjList(topList *ast.ObjectList) (map[string]StringSet, error)
 
 		resourceList := item.Val.(*ast.ObjectType).List
 		for _, rolesItem := range resourceList.Items {
-			key := rolesItem.Keys[0].Token.Text
+			key := rolesItem.Keys[0].Token.Value().(string)
```
jefferai: Is there ever a chance this might be `nil` (or not a string)? If so I'd prefer if you split this into two lines, something like 
```keyRaw := rolesItems.Keys[0].Token.Value()
key, ok := keyRaw.(string)
if !ok {
// Return some kind of error
}
```

emilymye: Done!

plugin/util/parse_bindings.go
```diff
@@ -88,12 +88,15 @@ func parseBindingObjList(topList *ast.ObjectList) (map[string]StringSet, error)
 
 		resourceList := item.Val.(*ast.ObjectType).List
 		for _, rolesItem := range resourceList.Items {
-			key := rolesItem.Keys[0].Token.Text
-			switch key {
+			if len(rolesItem.Keys) == 0 || rolesItem.Keys[0].Token.Value() == nil {
```
sethvargo: I think we also need to check if `rolesItem.Keys[0].Token` isn't nil before calling `.Value()`

plugin/util/parse_bindings.go
```diff
@@ -88,12 +88,15 @@ func parseBindingObjList(topList *ast.ObjectList) (map[string]StringSet, error)
...
+				merr = multierror.Append(merr, fmt.Errorf("invalid nil item under key %q in resource %q (line %d)", key, resourceName, rolesItem.Assign.Line))
+			}
+			roleKey := rolesItem.Keys[0].Token.Value().(string)
```
sethvargo: Not sure if `Value()` can be anything other than a string, but I would feel more comfortable if we did:

```golang
roleKey, ok := rolesItem.Keys[0].Token.Value().(string)
if !ok {
  // handle error
}
```

jefferai: Yeah that's exactly what I asked for too :-)

emilymye: my bad poor reading comprehension

plugin/util/parse_bindings.go
```diff
@@ -74,7 +74,12 @@ func parseBindingObjList(topList *ast.ObjectList) (map[string]StringSet, error)
 			continue
 		}
 
-		key := item.Keys[0].Token.Value().(string)
+		keyRaw := item.Keys[0].Token.Value()
```
jefferai: Probably we also need a nil check on keyRaw -- since it's an interface, it *could* be nil.

plugin/util/parse_bindings.go
jefferai: This seems to be the only issue left.

plugin/util/parse_bindings.go
```diff
@@ -70,30 +70,47 @@ func parseBindingObjList(topList *ast.ObjectList) (map[string]StringSet, error)
...
 
 		resourceList := item.Val.(*ast.ObjectType).List
 		for _, rolesItem := range resourceList.Items {
-			key := rolesItem.Keys[0].Token.Text
-			switch key {
+			if len(rolesItem.Keys) == 0 || rolesItem.Keys[0].Token.Value() == nil {
+				merr = multierror.Append(merr, fmt.Errorf("invalid nil item under key %q in resource %q (line %d)", key, resourceName, rolesItem.Assign.Line))
+				continue
+			}
+			roleKeyRaw := rolesItem.Keys[0].Token.Value()
```
jefferai: Oh except for a nil check here.

plugin/util/parse_bindings.go
```diff
@@ -69,68 +69,122 @@ func parseBindingObjList(topList *ast.ObjectList) (map[string]StringSet, error)
...
+			parsedRoles.Add(role)
+		}
+	}
+	return merr.ErrorOrNil()
 }
 
-func parseRoles(item *ast.ObjectItem, roleSet StringSet, merr *multierror.Error) {
-	lst, ok := item.Val.(*ast.ListType)
+func parseRole(parent *ast.ObjectItem, roleNode ast.Node) (string, error) {
+	roleLitType, ok := roleNode.(*ast.LiteralType)
```
jefferai: ast.Node is an interface type which means it can be nil, so it should be nil-checked.

sethvargo: @emilymye did you see Jeff's comment?

emilymye: @jefferai @sethvargo sorry fixing now - was on bug duty for TF last week so catching up on previous work now

PR #33, opened 2019-04-02
Cache clients

sethvargo:
Similar to what we did in the auth method

There's optimizations for more sharing between the libraries, but I think we should take the quick win for now.


sethvargo:
@jefferai @emilymye

PR #36, opened 2019-05-16
Remove redundant service_account/key verification before token generation

emilymye:
Remove .get() calls agains IAM service accounts/keys to verify token roleset. 

- The token generation step fails immediately anyways if the account/key are not valid, so these would mostly be for redundancy/better error messages. 

- Context: User requested 20x quota increase for read QPS to IAM and we'd like to avoid that. The OAuth token endpoint is not quota-limited so don't add limits from IAM API. 

Flaky test improvements.

- In general, IAM API can have delayed propagation, so retry the initial token check, ignore clean-up warnings from delayed deletes, etc.

### DIFF
--- a/plugin/secrets_access_token.go
+++ b/plugin/secrets_access_token.go
@@ -11,7 +11,6 @@ import (
 	"github.com/hashicorp/vault/sdk/logical"
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
-	"google.golang.org/api/iam/v1"
 )
 
 func pathSecretAccessToken(b *backend) *framework.Path {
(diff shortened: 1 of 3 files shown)


emilymye:
@sethvargo

plugin/secrets_access_token.go
```diff
@@ -58,19 +57,13 @@ func (b *backend) secretAccessTokenResponse(ctx context.Context, s logical.Stora
...
-	}
-
 	if rs.TokenGen == nil || rs.TokenGen.KeyName == "" {
 		return logical.ErrorResponse(fmt.Sprintf("invalid role set has no service account key, must be updated (path roleset/%s/rotate-key) before generating new secrets", rs.Name)), nil
 	}
 
 	token, err := rs.TokenGen.getAccessToken(ctx, iamC)
 	if err != nil {
-		return logical.ErrorResponse(fmt.Sprintf("could not generate token: %v", err)), nil
+		return logical.ErrorResponse(fmt.Sprintf("unable to generate token - make sure your roleset service account and key are still valid: %v", err)), nil
```
kalafut: ErrorResponse now has format string support.

```suggestion
		return logical.ErrorResponse("unable to generate token - make sure your roleset service account and key are still valid: %v", err), nil
```

plugin/secrets_access_token.go
```diff
@@ -83,14 +76,6 @@ func (b *backend) secretAccessTokenResponse(ctx context.Context, s logical.Stora
 }
 
 func (tg *TokenGenerator) getAccessToken(ctx context.Context, iamAdmin *iam.Service) (*oauth2.Token, error) {
```
kalafut: `iamAdmin` parameter can be removed.

kalafut:
@emilymye Some minor comments, otherwise LGTM.

emilymye:
@kalafut updated!

PR #39, opened 2019-05-24
Fix error handling to allow 403 for keys

### LINKED ISSUE #37: Issue disabling gcp secrets
I'm having issues disabling the gcp secrets engine, its trying to delete a key it issued that no longer exists.  Is there some way to force the delete from vault secrets disable gcp?

Full error when trying to disable:

There was an error disabling the gcp secrets engine at gcp/: failed to revoke "some-key" (1 / 2): failed to revoke entry: resp: &logical.Response{Secret:<nil>, Auth:<nil>, Data:map[string]interface {}{"error":"unable to delete service account key: googleapi: Error 403: Permission iam.serviceAccountKeys.delete is required to perform this operation on service account key projects/some-project/serviceAccounts/service-account-email/keys/key_id., forbidden"}, Redirect:"", Warnings:[]string(nil), WrapInfo:(*wrapping.ResponseWrapInfo)(nil)} err: <nil>.

emilymye:
Fixes #37.

### DIFF
--- a/plugin/secrets_access_token.go
+++ b/plugin/secrets_access_token.go
@@ -51,11 +51,6 @@ func (b *backend) pathAccessToken(ctx context.Context, req *logical.Request, d *
 }
 
 func (b *backend) secretAccessTokenResponse(ctx context.Context, s logical.Storage, rs *RoleSet) (*logical.Response, error) {
-	iamC, err := b.IAMClient(s)
-	if err != nil {
-		return nil, errwrap.Wrapf("could not create IAM Admin client: {{err}}", err)
-	}
-
 	if rs.TokenGen == nil || rs.TokenGen.KeyName == "" {
 		return logical.ErrorResponse("invalid role set has no service account key, must be updated (path roleset/%s/rotate-key) before generating new secrets", rs.Name), nil
 	}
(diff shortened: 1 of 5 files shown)


plugin/secrets_access_token.go
```diff
@@ -51,11 +51,6 @@ func (b *backend) pathAccessToken(ctx context.Context, req *logical.Request, d *
 }
 
 func (b *backend) secretAccessTokenResponse(ctx context.Context, s logical.Storage, rs *RoleSet) (*logical.Response, error) {
-	iamC, err := b.IAMClient(s)
```
emilymye: iamC is no longer used, so it's safe to delete. Leftover from #36.

emilymye:
cc @sethvargo for review?

jefferai:
You said in some cases it is a 403, is the behavior the same regardless of whether it's a 403/404 and should just check for both?

emilymye:
@jefferai not sure I understand the previous comment but yes, we're essentially just checking that it's one or the other. I guess this does mask if the Vault credentials also starts getting a 403 on the key if its credentials on the key are revoked, but I'm not sure if it makes sense for use to throw an error then. Maybe a warning?

jefferai:
What I was really getting at is, is there a reason not to just always check for either? I don't know those exact rules for when it's one or the other but if that behavior can change then this might prove fragile. If it's clear what those cases are then it's probably fine.

jefferai:
Merging.

emilymye:
@jefferai sorry for late reply (not that i think it will change anything) - we are actually checking for either, so yes? :) The PR keeps 404 for serviceAccount.get but checks for 403/404 for keys, since I'm fairly sure we won't run into the same issue for serviceAccounts (because Terraform doesn't) and I don't want to mask the actual permission error for that.

PR #44, opened 2019-07-01
Don't rotate service account for exactly same roleset bindings

### LINKED ISSUE #43: Re-Apply of the same roleset causes Vault to delete and recreate the google Service Account
Whenever a Roleset configuration , with no changes, is re-applied to Vault it forces it to destroy and recreate the Google Service account invalidating any outstanding Key for the service account that was deleted.

The reason why this is an issue, at least for me, is that i am using a [kubernetes operator](https://github.com/banzaicloud/bank-vaults) to apply the Vault configurations including configuring the gcp secret backend.<br>
In kubernetes whenever the `configurator pod` is recreated ( due to pod being killed or upgrading the operator itself ) then the coniguration will be re-applied and the service account on google deleted and recreated.

I have spent a bit of time trying to understand if i could handle this on the operator side but i am having big issues matching the format returned by vault when reading the roleset to what the `hcl.Decode` library from the configuration string returns

for example, this is what i get by reading a roleset binding from vault 
```
map[string]interface {}{"//cloudresourcemanager.googleapis.com/projects/mintel-k8s-test1-3134ed5d":[]interface {}{"roles/viewer"}}
```

vs what i get by hcl.Decode on the source configuration for that roleset
```
map[string]interface {}{"resource":[]map[string]interface {}{map[string]interface {}{"//cloudresourcemanager.googleapis.com/projects/mintel-k8s-test1-
3134ed5d":[]map[string]interface {}{map[string]interface {}{"roles":[]interface {}{"roles/viewer"}}}}}}
```

I wanted to understand if this behavior from Vault is expected and desired behavior before spending more time trying to have the operator understand if there is any difference between the already configured roleset and the one in the source configuration.

Most ( all ? ) the other backends in Vault seem to handle reconfiguration much better than this one.

## How to reproduce with Vault Cli

Start Vault ( c.json should be the google credentials json file )
```
docker run --name vault --rm -t -i -e GOOGLE_APPLICATION_CREDENTIALS="/p/c.json" -e SKIP_SETCAP=true -e VAULT_DEV_ROOT_TOKEN_ID=227e1cce --net=host -v /dev/shm/:/p vault:1.1.3 server -dev -dev-listen-address=0.0.0.0:8200
```

configure the GCP Secret Backend
```
export VAULT_ADDR="http://localhost:8200"
export VAULT_TOKEN="227e1cce"

vault secrets enable gcp
```

Create the test-key roleset
```
vault write gcp/roleset/test-key \
    project="mintel-k8s-test1-3134ed5d" \
    bindings=-<<EOF
      resource "//cloudresourcemanager.googleapis.com/projects/mintel-k8s-test1-3134ed5d" {
        roles = ["roles/viewer"]
      }
EOF

vault read gcp/roleset/test-key -format=json | jq .data.service_account_email
"vaulttest-key-1561556571@mintel-k8s-test1-3134ed5d.iam.gserviceaccount.com"
```

Reapply the same test-key roleset and confirm change of service account user email
```
vault write gcp/roleset/test-key \
    project="mintel-k8s-test1-3134ed5d" \
    bindings=-<<EOF
      resource "//cloudresourcemanager.googleapis.com/projects/mintel-k8s-test1-3134ed5d" {
        roles = ["roles/viewer"]
      }
EOF

vault read gcp/roleset/test-key -format=json | jq .data.service_account_email
"vaulttest-key-1561556620@mintel-k8s-test1-3134ed5d.iam.gserviceaccount.com"
```

emilymye:
Fixes #43

### DIFF
--- a/plugin/path_role_set.go
+++ b/plugin/path_role_set.go
@@ -338,7 +338,9 @@ func (b *backend) pathRoleSetCreateUpdate(ctx context.Context, req *logical.Requ
 		return logical.ErrorResponse("bindings are required for new role set"), nil
 	}
 
-	if !newBindings {
+	// If no new bindings or new bindings are exactly same as old bindings,
+	// just update the role set without rotating service account.
+	if !newBindings || rs.bindingHash() == getStringHash(bRaw.(string)) {
 		// Just save role with updated metadata:
 		if err := rs.save(ctx, req.Storage); err != nil {
 			return logical.ErrorResponse(err.Error()), nil
@@ -356,6 +358,7 @@ func (b *backend) pathRoleSetCreateUpdate(ctx context.Context, req *logical.Requ
 		return logical.ErrorResponse("unable to parse any bindings from given bindings HCL"), nil
 	}
 	rs.RawBindings = bRaw.(string)
+
 	updateWarns, err := b.saveRoleSetWithNewAccount(ctx, req.Storage, rs, project, bindings, scopes)
 	if updateWarns != nil {
 		warnings = append(warnings, updateWarns...)
--- a/plugin/role_set.go
+++ b/plugin/role_set.go
@@ -88,8 +88,7 @@ func (rs *RoleSet) save(ctx context.Context, s logical.Storage) error {
 }
 
 func (rs *RoleSet) bindingHash() string {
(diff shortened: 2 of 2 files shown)


primeroz:
Great thanks ! 

should i just follow the Vault release / changelog to check when this makes it into vault ?

PR #49, opened 2019-10-02
Refactor roleset

emilymye:
As a precursor to adding two new fields, I wanted to refactor the roleset code.

Some larger changes:
- Separated `pathRoleSetCreateUpdate` into two paths: this will be very helpful with more fields and the code was starting to get confusing.
- Allow empty `bindings` (param must still be given, just as an empty string, to differentiate in update between "i want to set to none" and "i don't want to change it")
- Code no longer try to remove WALs that are not being used (i.e. we successfully deleted an account so we removed the WALs for rolling it back). The WAL callbacks check to make sure a roleset in storage isn't still using a resource (account/key/binding), so there's no harm in having them other than maybe more calls to storage, and the removal code required specific order requirements, which we already have enough of. 

Also:
- removed logical.ErrorResponse(fmt.Sprintf(...)) usage for new ErrorResponse()
- changed log level handling for tests

### DIFF
--- a/plugin/backend_test.go
+++ b/plugin/backend_test.go
@@ -2,10 +2,12 @@ package gcpsecrets
 
 import (
 	"context"
+	"github.com/hashicorp/vault/sdk/helper/logging"
+	"os"
 	"testing"
 	"time"
 
-	hclog "github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-hclog"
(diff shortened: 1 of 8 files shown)


emilymye:
@kalafut at this point i might try to reduce the refactoring so it's less of a burden - please hold on PR!

kalafut:
>please hold on PR!

If I must... 😉

PR #53, opened 2019-11-14
Add support for rotating root keys

sethvargo:
This adds a new path `/config/rotate-root` that rotates the underlying service account key so only Vault knows it's value.


sethvargo:
@emilymye

sethvargo:
@kalafut

plugin/path_config_rotate_root.go
```diff
@@ -0,0 +1,128 @@
...
+	// Get the current configuration
+	cfg, err := getConfig(ctx, req.Storage)
+	if err != nil {
+		return nil, err
+	}
+	if cfg == nil {
+		return nil, fmt.Errorf("no configuration")
+	}
+	if cfg.CredentialsRaw == "" {
+		return nil, fmt.Errorf("configuration does not have credentials")
```
emilymye: maybe use error message saying this path only applies to config with explicitly-set JSON credentials? so ADC users aren't confused

sethvargo: Done

go.mod
```diff
@@ -1,9 +1,10 @@
 module github.com/hashicorp/vault-plugin-secrets-gcp
 
-go 1.12
+go 1.13
+
+replace git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
```
kalafut: At least in the main repo we've pulled this out:  https://github.com/hashicorp/vault/commit/fa245e6a1db4bd8b8328c6c1a9850269078a5da9

plugin/path_config_rotate_root_test.go
```diff
@@ -0,0 +1,194 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Create http client
+		clientCtx := context.WithValue(ctx, oauth2.HTTPClient, cleanhttp.DefaultClient())
+		client := oauth2.NewClient(clientCtx, creds.TokenSource)
+
+		// Create IAM client
+		iamAdmin, err := iam.New(client)
```
kalafut: Deprecation warning:

> iam.New is deprecated: please use NewService instead. To provide a custom HTTP client, use option.WithHTTPClient. If you are using google.golang.org/api/googleapis/transport.APIKey, use option.WithAPIKey with NewService instead.

This warning hits a few other areas too, but since this is new code it would be good to use the current version.

plugin/path_config_rotate_root.go
```diff
@@ -0,0 +1,130 @@
...
+	// Generate a new service account key
+	iamAdmin, err := b.IAMAdminClient(req.Storage)
+	if err != nil {
+		return nil, errwrap.Wrapf("failed to create iam client: {{err}}", err)
+	}
+
+	saName := "projects/-/serviceAccounts/" + creds.ClientEmail
+	newKey, err := iamAdmin.Projects.ServiceAccounts.Keys.
+		Create(saName, &iam.CreateServiceAccountKeyRequest{
+			KeyAlgorithm:   "KEY_ALG_RSA_2048",
```
kalafut: There existing constants `keyAlgorithmRSA2k` and `privateKeyTypeJson` for these values.

plugin/path_config_rotate_root.go
```diff
@@ -0,0 +1,130 @@
...
+		creds.PrivateKeyId)
+	if _, err := iamAdmin.Projects.ServiceAccounts.Keys.
+		Delete(oldKeyName).
+		Context(ctx).
+		Do(); err != nil {
+		return nil, errwrap.Wrapf("failed to delete old service account key: {{err}}", err)
+	}
+
+	// Clear caches to pick up the new credentials
+	b.ClearCaches()
```
kalafut: I feel this should probably go before the delete, otherwise on a delete fail we'll be in a inconsistent state between cache and store.

plugin/path_config_rotate_root.go
```diff
@@ -0,0 +1,130 @@
...
+	// Delete the old service account key
+	oldKeyName := fmt.Sprintf("projects/%s/serviceAccounts/%s/keys/%s",
+		creds.ProjectId,
+		creds.ClientEmail,
+		creds.PrivateKeyId)
+	if _, err := iamAdmin.Projects.ServiceAccounts.Keys.
+		Delete(oldKeyName).
+		Context(ctx).
+		Do(); err != nil {
+		return nil, errwrap.Wrapf("failed to delete old service account key: {{err}}", err)
```
kalafut: Should we instead log this as a warning and continue to return `"private_key_id": newCreds.PrivateKeyId,` since the updated config was already written?

go.mod
sethvargo: Oh nice. I'm not gonna question why it's suddenly fixed.

plugin/path_config_rotate_root.go
sethvargo: Done

plugin/path_config_rotate_root.go
sethvargo: I moved it to go before delete. I think we should return an error if the delete fails. Users don't look at warnings (usually).

plugin/path_config_rotate_root_test.go
sethvargo: I'm aware. It's not really feasible to update to `NewService` since it directly conflicts with `New`. I'd have to update all the files and change around some internals because of the new constructor. I'm happy to do that, but I'd prefer to take on a "fix all the warnings" PR instead of merging it with this one. Is that okay with you?

plugin/path_config_rotate_root.go
sethvargo: I've updated the error to include the old and new key ids. I really think we should return an error though. Users are likely to ignore errors.

sethvargo:
@kalafut updated! Neither Emily nor I have merge permissions though.

plugin/path_config_rotate_root_test.go
kalafut: Sounds good.

plugin/path_config_rotate_root.go
kalafut: 👍  Returning an error is fine now that it clearly indicates that the rotation did occur.

plugin/path_config_rotate_root_test.go
```diff
@@ -0,0 +1,194 @@
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// Create a new key, since this endpoint revokes the old key
+		saName := "projects/-/serviceAccounts/" + parsedCreds.ClientEmail
+		newKey, err := iamAdmin.Projects.ServiceAccounts.Keys.
+			Create(saName, &iam.CreateServiceAccountKeyRequest{
+				KeyAlgorithm:   "KEY_ALG_RSA_2048",
```
kalafut: Could you update these with the constants too?

kalafut:
@sethvargo It LGTM except for one little thing I missed in the test. Otherwise it good merge.

plugin/path_config_rotate_root_test.go
sethvargo: Done!

sethvargo:
@kalafut updated

kalafut:
@sethvargo Just realized that this will error on `make test` if the user doesn't have GOOGLE_CREDENTIALS set/present. One fix is to add to the start of the [rotate test](https://github.com/hashicorp/vault-plugin-secrets-gcp/pull/53/files#diff-f80ef5a8cd1852f0a117ab0d6eff00a6R99):
```
	if testing.Short() {
		t.Skip("skipping integration test (short)")
	}
```
I thought `GetTestCredentials` (which does this check) would be a drop in but the types are a little different. Your call...

sethvargo:
Good catch. I just added the quick skip for now

PR #54, opened 2019-11-15
Allow specifying a TTL on a key

sethvargo:


plugin/secrets_service_account_key.go
```diff
@@ -195,6 +201,22 @@ func (b *backend) getSecretKey(ctx context.Context, s logical.Storage, rs *RoleS
...
+	}
+
+	maxTTL := int(cfg.MaxTTL.Seconds())
+	if maxTTL <= 0 {
+		maxTTL = int(b.System().MaxLeaseTTL().Seconds())
+	}
+
+	if ttl > maxTTL {
+		ttl = maxTTL
+	}
```
emilymye: should we add a warning to the response?

sethvargo: I thought about it, but none of the other plugins push out a warning, and the response itself includes the computed TTL.

emilymye: 👍 i'm fine either way, just thought it was nicer than ignoring the TTL given

kalafut: I don't _think_ these parts are needed.

- Vault will fall back to system Default and Max TTLs if the return secret TTLs  are <= 0
- Vault will cap the TTL at the MaxTTL automatically
- If we let Vault do that limiting (i.e. remove 216-218), then you'll get a warning response for free:
```
WARNING! The following warnings were returned from Vault:

  * TTL of "11h0m0s" exceeded the effective max_ttl of "10h0m0s"; TTL value is
  capped accordingly
```

sethvargo: I verified Jim's comment below. Turns out we don't need any of this :smile:

sethvargo: Yup you're right. I just verified this. Updated.

sethvargo:
@kalafut updated - thanks for the pointer. Just as a headers up, neither Emily nor I have permission to merge changes.

kalafut:
@sethvargo I've got some other docs changes going in and will make updates re: this PR unless you've already got something in the works.

sethvargo:
Hey @kalafut I have not started docs - was gonna wait for merge. Happy to defer to you. Let me know if you need help/review 😄

kalafut:
@sethvargo I'll grab this one. Rotate root might (?) warrant a section and some more words, so some text there would be much appreciated.

sethvargo:
@kalafut can you CC me on your PR in core and I can drop some words in :)

kalafut:
@sethvargo Sure. Might get to those docs today...

PR #58, opened 2019-11-19
Update google and vault deps

jefferai:


jefferai:
Note: please merge, not squash, when committing.

emilymye:
got an error from github.com/hashicorp/go-retryablehttp/client.go: c.HTTPClient.CloseIdleConnections undefined (type *http.Client has no field or method CloseIdleConnections)

PR #62, opened 2019-11-21
Refactor roleset into utils

emilymye:


plugin/path_role_set.go
```diff
@@ -6,123 +6,113 @@ import (
...
-				},
+		HelpSynopsis:    pathRoleSetHelpSyn,
+		HelpDescription: pathRoleSetHelpDesc,
+	}
+}
+
+func pathRoleSetList(b *backend) *framework.Path {
+	// Paths for listing role sets
+	return &framework.Path{
+		Pattern: "rolesets?/?",
```
sethvargo: Is the `?` after `rolesets` intended? `rolesets?` seems weird

plugin/role_set_utils.go
```diff
@@ -0,0 +1,161 @@
...
+		accountId:   rs.AccountId,
+		rolesetName: rs.Name,
+		bindings:    rs.Bindings,
+		tokenGen:    rs.TokenGen,
+	}
+}
+
+// addWalsForRoleSetResources creates WALs to clean up a roleset's service account, bindings, and a key if needed.
+func (b *backend) addWalsForRoleSetResources(ctx context.Context, req *logical.Request, boundResources *roleSetResources) error {
+	if boundResources == nil {
```
sethvargo: Should we behave the same if `boundResources` is empty? If you change this to a length check, it'll handle both.

plugin/role_set_utils.go
```diff
@@ -0,0 +1,161 @@
...
+		b.Logger().Debug("skip deletion for nil roleset resources")
+		return nil
+	}
+
+	httpC, err := b.HTTPClient(req.Storage)
+	if err != nil {
+		return []string{err.Error()}
+	}
+
+	iamAdmin, err := iam.NewService(ctx, option.WithHTTPClient(httpC))
```
sethvargo: You should be able to call [`b.IAMAdminClient()` without needing the http client](https://github.com/hashicorp/vault-plugin-secrets-gcp/blob/9c1dd008b73a781f4954215084a1b3d104d88a81/plugin/backend.go#L91)

plugin/role_set_utils.go
```diff
@@ -0,0 +1,161 @@
...
+	if err != nil {
+		return errwrap.Wrapf("unable to create WAL entry to clean up service account key: {{err}}", err)
+	}
+	return nil
+}
+
+// tryDeleteRoleSetResources tries to delete GCP resources previously managed by a roleset.
+// This assumes that deletion of these resources will already be guaranteed by WALs and will return errors
+// as a list of warnings instead.
+func (b *backend) tryDeleteRoleSetResources(ctx context.Context, req *logical.Request, boundResources *roleSetResources) []string {
```
sethvargo: I think we should return `(err error, warnings []string)` here. It feels weird to swallow up "failed to create HTTP client" as a warning.

plugin/role_set_utils.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	return nil
+}
+
+func emailForServiceAccountName(project, accountName string) string {
+	return fmt.Sprintf(serviceAccountEmailTemplate, accountName, project)
+}
+
+func generateAccountNameForRoleSet(rsName string) (name string) {
```
sethvargo: I realize it's private, but can you document this with the expected input and expected output? There's a lot of string magic so have 1-2 examples of "here's what this function will output" could be helpful

plugin/path_role_set.go
emilymye: yup - we had two paths for roleset/? and rolesets/? so might as well match 0-1 of the s

plugin/role_set_utils.go
emilymye: welp this is confusing - boundResources is `*rolesetResources`, a new object I created to wrap:

- a service account
- ResourceBindings
- TokenGen (key)

It's handier than passing around a whole roleset and allows me to remove some of the confusion where it wasn't clear which updated fields were updated in the RoleSet vs being passed into the function

emilymye: I am very open to naming help

plugin/role_set_utils.go
emilymye: I'll switch it so it uses the cached version, I think I used this because we needed the HTTP client anyways for the iamHandle. 

Side question - is HTTP client is the biggest thing we wanted to cache?

plugin/role_set_utils.go
emilymye: I do see why we might want to elevate the warning to an error, but I'm not sure I agree with the mixed approach.

Feedback is welcome. If we're trying to say "We succeeded in deleting/updating the roleset (intended operation) but we encountered an error while trying to clean up old resources - this will be covered by WAL", would it be better to 

- warn on any error - the actual operation succeeded and the other resources will get cleaned up eventually
- return the error - the actual roleset was created but we got an error, so if you're worried about any time period where there is unmanaged resources, you should go delete them
- err on the big things (can't create clients) but warn if we can't clean up the other errors.  

I do see why we might want to elevate the warning to an error, but I'm not sure I agree with the mixed approach, especially since the actual operation succeeded, which means recently the HTTP client was creatable

plugin/role_set_utils.go
```diff
@@ -0,0 +1,165 @@
+package gcpsecrets
```
emilymye: as a note, ideally i'll move the rest of the shared utils from role_set into role_set utils so roleset only includes the Roleset related types and methods on those types

plugin/path_role_set.go
sethvargo: Ack

plugin/role_set_utils.go
sethvargo: Heh that's fun. Umm I think this is fine then.

plugin/role_set_utils.go
sethvargo: It's Russian doll cached. So IAMClient creates and caches an HTTP client too.

emilymye:
@kalafut sorry, please hold off on merging/reviewing this - I realized something I removed (deleting WALs that aren't needed anymore) should stay in to reduce the number of GCP calls, and I've been working on the alternative (static role) path so my refactors have changed a bit - I'll push an update to this PR and will have a PR for the new path as well

sorry for so much churn, and thanks for all the reviews!

plugin/secrets_service_account_key.go
```diff
@@ -174,26 +174,17 @@ func (b *backend) secretKeyRevoke(ctx context.Context, req *logical.Request, d *
...
+	if rs.AccountId == nil {
+		return nil, fmt.Errorf("unable to create secret (service account key), roleset has nil account ID")
 	}
 
 	iamC, err := b.IAMAdminClient(s)
 	if err != nil {
 		return nil, errwrap.Wrapf("could not create IAM Admin client: {{err}}", err)
 	}
 
-	account, err := rs.getServiceAccount(iamC)
```
emilymye: Removing this - we did this for access tokens because it was potentially killing user's IAM quota. While I don't think the same issue happens with service account keys (since hard limit of 10 per service account), there's no real reason to have this check. keys.Create will 404 if the service account doesn't exist so we don't gain much error message

plugin/secrets_service_account_key.go
```diff
@@ -174,26 +174,17 @@ func (b *backend) secretKeyRevoke(ctx context.Context, req *logical.Request, d *
 }
 
 func (b *backend) getSecretKey(ctx context.Context, s logical.Storage, rs *RoleSet, keyType, keyAlgorithm string, ttl int) (*logical.Response, error) {
-	cfg, err := getConfig(ctx, s)
```
emilymye: removing unused cfg, it is left over from when we used to look up TTL from config

plugin/role_set.go
```diff
@@ -91,316 +101,256 @@ func (rs *RoleSet) bindingHash() string {
...
+		return nil, errwrap.Wrapf(fmt.Sprintf("could not find service account %q: {{err}}", accountId.ResourceName()), err)
 	} else if account == nil {
-		return nil, fmt.Errorf("roleset service account was removed - role set must be updated (path roleset/%s/rotate) before generating new secrets", rs.Name)
+		return nil, fmt.Errorf("could not find service account %q: {{err}}", err)
 	}
 
 	return account, nil
 }
 
-type ResourceBindings map[string]util.StringSet
```
emilymye: these got moved to gcp_account_resources, since they represent GCP resources

plugin/role_set.go
```diff
@@ -91,316 +101,256 @@ func (rs *RoleSet) bindingHash() string {
...
-	}
-	return out
-}
-
-type TokenGenerator struct {
-	KeyName    string
-	B64KeyJSON string
+// saveRoleSetWithNewAccount rotates the role set service account. This includes creating a new service account with
+// a new name and deleting the old service account, updating keys or bindings as required.
+func (b *backend) saveRoleSetWithNewAccount(ctx context.Context, req *logical.Request, rs *RoleSet, project string, newBinds ResourceBindings, scopes []string) (warnings []string, err error) {
```
emilymye: this diff is truly terrible, I am sorry

plugin/gcp_account_resources.go
```diff
@@ -0,0 +1,135 @@
...
+func (rb ResourceBindings) asOutput() map[string][]string {
+	out := make(map[string][]string)
+	for k, v := range rb {
+		out[k] = v.ToSlice()
+	}
+	return out
+}
+
+func getStringHash(bindingsRaw string) string {
+	ssum := sha256.Sum256([]byte(bindingsRaw)[:])
```
kalafut:
```suggestion
	ssum := sha256.Sum256([]byte(bindingsRaw))
```

plugin/gcp_account_resources.go
```diff
@@ -0,0 +1,135 @@
...
+	// TokenGenerator wraps the service account key and params required to create access tokens.
+	TokenGenerator struct {
+		KeyName    string
+		B64KeyJSON string
+		Scopes     []string
+	}
+)
+
+func (rb ResourceBindings) asOutput() map[string][]string {
+	out := make(map[string][]string)
```
kalafut:
```suggestion
	out := make(map[string][]string, len(rb))
```

plugin/gcp_account_resources.go
```diff
@@ -0,0 +1,135 @@
...
+		if err != nil {
+			return nil
+		}
+
+		b.Logger().Debug("got IAM policy for resource name", "name", resourceName)
+		changed, newP := p.AddBindings(&iamutil.PolicyDelta{
+			Roles: roles,
+			Email: saEmail,
+		})
+		if !changed || newP == nil {
```
kalafut: Can `newP` ever be `nil`?

plugin/path_role_set.go
```diff
@@ -270,6 +215,9 @@ func (b *backend) pathRoleSetCreateUpdate(ctx context.Context, req *logical.Requ
 	}
 	name := nameRaw.(string)
```
kalafut: Since `name` is a path parameter, it's presence will be guaranteed by the framework (i.e. the `if !ok {...}` will never be hit). So you can optionally do: `name := d.Get("name").(string)`

plugin/rollback.go
```diff
@@ -3,13 +3,13 @@ package gcpsecrets
 import (
 	"context"
 	"fmt"
+	"github.com/hashicorp/vault/sdk/framework"
```
kalafut: Run `goimports`

plugin/role_set.go
```diff
@@ -91,316 +101,256 @@ func (rs *RoleSet) bindingHash() string {
 	return getStringHash(rs.RawBindings)
 }
 
-func (rs *RoleSet) getServiceAccount(iamAdmin *iam.Service) (*iam.ServiceAccount, error) {
-	if rs.AccountId == nil {
-		return nil, fmt.Errorf("role set '%s' is invalid, has no associated service account", rs.Name)
+// getServiceAccount gets the
```
kalafut: gets the ...

plugin/role_set.go
```diff
@@ -91,316 +101,256 @@ func (rs *RoleSet) bindingHash() string {
...
 	}
 
-	account, err := iamAdmin.Projects.ServiceAccounts.Get(rs.AccountId.ResourceName()).Do()
+	account, err := iamAdmin.Projects.ServiceAccounts.Get(accountId.ResourceName()).Do()
 	if err != nil {
-		return nil, fmt.Errorf("could not find service account: %v. If account was deleted, role set must be updated (write to roleset/%s/rotate) before generating new secrets", err, rs.Name)
+		return nil, errwrap.Wrapf(fmt.Sprintf("could not find service account %q: {{err}}", accountId.ResourceName()), err)
 	} else if account == nil {
-		return nil, fmt.Errorf("roleset service account was removed - role set must be updated (path roleset/%s/rotate) before generating new secrets", rs.Name)
+		return nil, fmt.Errorf("could not find service account %q: {{err}}", err)
```
kalafut: The `%q` and `{{err}}` here aren't quite right.

plugin/role_set.go
```diff
@@ -91,316 +101,256 @@ func (rs *RoleSet) bindingHash() string {
...
 
-func roleSetServiceAccountName(rsName string) (name string) {
+// generateAccountNameForRoleSet returns a new random name for a Vault service account based off roleset name and time.
+// Note this is the name rather than the full email (i.e. string before @)
+//
+// As an example, for roleset "my-role" this returns `vaultmy-role-1234613`
+//
+func generateAccountNameForRoleSet(rsName string) (name string) {
 	// Sanitize role name
 	reg := regexp.MustCompile("[^a-zA-Z0-9-]+")
```
kalafut: This would be better as top level variable, compiled only once (and panicking immediately instead of in the request time path).

plugin/role_set_utils.go
kalafut: No strong opinions from me on this one, but I'd be fine if these were just treated as errors. Granted, the user impact is low, but we treat the inability to create a roleset as an error, so problems deleted them (even if for clean-up) seem error-worthy. If desired you could also work in go-multierror, which already used elsewhere.

plugin/gcp_account_resources.go
emilymye: yes/no (changed would be false, so would catch) - i don't see any harm in keeping the check though in case?

kalafut: It's OK as is. Mainly was curious.

Issue #72, opened 2020-01-06
Unable to generate a service account key with a specific TTL


mpailloncy:
I'd like the ability to generate Service Account key with a specific TTL (not the one by default or the one configured in gcp/config).

Following the [documentation](https://www.vaultproject.io/api/secret/gcp/index.html#generate-secret-iam-service-account-creds-service-account-key), I tried to execute:

`curl --header "X-Vault-Token: XXXX" --request POST --data @payload.json 'https://<VAULT_ADDR>/v1/gcp/key/<ROLE_NAME>'`

With this payload:
```
{
  "roleset": "<ROLE_NAME>",
  "key_algorithm": "TYPE_GOOGLE_CREDENTIALS_FILE",
  "key_type": "KEY_ALG_RSA_2048",
  "ttl": "5m"
}
```

Even though the "ttl" is not listed as available parameter in the documentation, it looks like [it can be specified when looking at the code](https://github.com/hashicorp/vault-plugin-secrets-gcp/blob/master/plugin/secrets_service_account_key.go#L61-L64).

Anyways, I'm getting always the same result with or without the "ttl" parameter:

> {"errors":["1 error occurred:\n\t* internal error\n\n"]}

and this error appears in vault logs:

`[ERROR] core: failed to run existence check: error="roleset name is required"`

Looking at the code, it looks like the declared field name is `roleset` [inside the Fields schema](https://github.com/hashicorp/vault-plugin-secrets-gcp/blob/master/plugin/secrets_service_account_key.go#L47), while a field `name` is expected inside the corresponding [ExistenceCheck](https://github.com/hashicorp/vault-plugin-secrets-gcp/blob/556b7c55d6db11a34eef1bbf6bc455bedfecdaf6/plugin/path_role_set.go#L124-L127).

PR #73, opened 2020-01-06
Fix roleset field name for service account key generation endpoint

mpailloncy:
This field name was "roleset", while [the corresponding ExistingCheck](https://github.com/hashicorp/vault-plugin-secrets-gcp/blob/556b7c55d6db11a34eef1bbf6bc455bedfecdaf6/plugin/path_role_set.go#L124-L127) is looking at a field named "name".
That causes any POST request performed against this endpoint to fail with error "core: failed to run existence check: error="roleset name is required" as [this condition](https://github.com/hashicorp/vault-plugin-secrets-gcp/blob/556b7c55d6db11a34eef1bbf6bc455bedfecdaf6/vendor/github.com/hashicorp/vault/sdk/framework/field_data.go#L106-L109) always return false because fields names mismatch.

Should fix https://github.com/hashicorp/vault-plugin-secrets-gcp/issues/72

### DIFF
--- a/plugin/secrets_service_account_key.go
+++ b/plugin/secrets_service_account_key.go
@@ -44,7 +44,7 @@ func pathSecretServiceAccountKey(b *backend) *framework.Path {
 	return &framework.Path{
 		Pattern: fmt.Sprintf("key/%s", framework.GenericNameRegex("roleset")),
 		Fields: map[string]*framework.FieldSchema{
-			"roleset": {
+			"name": {
 				Type:        framework.TypeString,
 				Description: "Required. Name of the role set.",
 			},


plugin/secrets_service_account_key.go
```diff
@@ -44,7 +44,7 @@ func pathSecretServiceAccountKey(b *backend) *framework.Path {
 	return &framework.Path{
 		Pattern: fmt.Sprintf("key/%s", framework.GenericNameRegex("roleset")),
```
emilymye: This also needs to be changed back to name then

emilymye:
Closing this PR as #76 fixed this - thanks for calling attention to this bug!

PR #76, opened 2020-02-03
Change existence check to allow for custom field name for roleset

### LINKED ISSUE #72: Unable to generate a service account key with a specific TTL

emilymye:
Fixes #72, Supersedes #73 - figured fix should get in faster and there hasn't been any movement on that PR

### DIFF
--- a/plugin/path_role_set.go
+++ b/plugin/path_role_set.go
@@ -45,7 +45,7 @@ func pathRoleSet(b *backend) *framework.Path {
 				Description: `List of OAuth scopes to assign to credentials generated under this role set`,
 			},
 		},
-		ExistenceCheck: b.pathRoleSetExistenceCheck,
+		ExistenceCheck: b.pathRoleSetExistenceCheck("name"),
 		Operations: map[logical.Operation]framework.OperationHandler{
 			logical.DeleteOperation: &framework.PathOperation{
 				Callback: b.pathRoleSetDelete,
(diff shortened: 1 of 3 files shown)


plugin/path_role_set.go
```diff
@@ -120,18 +120,21 @@ func pathRoleSetRotateKey(b *backend) *framework.Path {
 	}
 }
 
-func (b *backend) pathRoleSetExistenceCheck(ctx context.Context, req *logical.Request, d *framework.FieldData) (bool, error) {
-	nameRaw, ok := d.GetOk("name")
-	if !ok {
-		return false, errors.New("roleset name is required")
-	}
+func (b *backend) pathRoleSetExistenceCheck(rolesetFieldName string) framework.ExistenceFunc {
```
sethvargo: Clever!

PR #78, opened 2020-02-05
feat: support BigQuery dataset ACLs in absence of IAM endpoints

### LINKED ISSUE #24: Roleset bindings should support BigQuery datasets
It is not possible to add bindings for a bigquery dataset resource.

Looks like the big query api does not provide `getIamPolicy`, `setIamPolicy` and `testIamPermissions`. See [update](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets/update) for [datasets](https://cloud.google.com/bigquery/docs/reference/rest/v2/#Datasets)

Is there any chance to add support for datasets, too?

onetwopunch:
Fixes #24 

Since BigQuery does not support the typical IAM set/get endpoints, we have to convert the ACL bindings to Policy and back again. This works because the dataset has a PATCH endpoint that means we won't overwrite any existing ACLs. Tested this out locally as well as added unit tests for the policy <> dataset conversion.

Also ran the following to generate the `iam_resources_generated.go` file in the absence of a Makefile entry for it. Happy to add one though if desired.:

```
cd plugin/iamutil
go run internal/generate_iam.go
```


emilymye:
I didn't quite think this through the first time around and some of my types are a bit muddled, so here's what I'll propose - I really would not like to do the isBigQueryResource() check but I'm willing to review as is, until I get in another PR that makes it easier for custom IAM code.
The way I originally envisioned doing this was implementing a custom BigQuery [IamResource](https://github.com/hashicorp/vault-plugin-secrets-gcp/blob/master/plugin/iamutil/iam_resource.go#L15) that then gets added to the Enabled IAM resources here: https://github.com/hashicorp/vault-plugin-secrets-gcp/blob/8286766357ef89f7a418997913edd9f0eec0d83c/plugin/iamutil/iam_resources_generated.go#L5

like 

```
func GetEnabledIamResources() GeneratedResources {
        combineResourceMaps(
            generatedIamResources,
            addBigQueryResources(generated),
        )
}

```
where then we could just do the custom BigQuery resource code that edits the IAM policy on the actual resource (as I believe they do).

onetwopunch:
@emilymye Could you clarify what you mean by "just do the custom BigQuery resource code that edits the IAM policy on the actual resource" ? From the documentation I could find on [BigQuery IAM at the dataset level](https://cloud.google.com/bigquery/docs/dataset-access-controls), it seems the ACLs are not in a separate API call but rather part of the [Dataset](https://cloud.google.com/bigquery/docs/reference/rest/v2/datasets#Dataset) resource. If I'm missing something please let me know, but I did at first attempt to go down that route you initially suggested in the issue, but since it would not only have required a new resource and resource list but also overrides on the IAM handle, it seemed like a lot of duplicate code. I'm happy to restructure this to fit the pattern you suggested, but I want to make sure we're on the same page of pathing Dataset object ACLs rather than IAM policy. Unless there is some other API I'm missing, I just couldn't find one. Please let me know your thoughts and thank you for the response :)

crazed:
This is going to make a project I just started much easier. Is there a chance we could add documentation example on how to setup the appropriate HCL definition for BigQuery?

onetwopunch:
@emilymye Refactored per our conversation. Tests are passing. Could you have another review

onetwopunch:
@crazed Sorry for the lag in response. The docs aren't actually in this repo, but I'll look into adding supporting documentation once this is merged. In the mean time, how you'd use this is:

```
vault write gcp/roleset/bq project=my-project secret_type=service_account_key bindings=@bindings.hcl
```
Where `bindings.hcl` looks like:

```hcl
resource "//bigquery.googleapis.com/projects/my-project/datasets/test_dataset" {
  roles = ["roles/bigquery.dataViewer"]
}
```

Hope this helps :)

onetwopunch:
@kalafut Would you mind having a look at this please?

plugin/iamutil/dataset_resource.go
```diff
@@ -0,0 +1,144 @@
...
+	ds := &Dataset{
+		Etag: p.Etag,
+	}
+
+	if p == nil {
+		return nil, errwrap.Wrap(errors.New("Policy cannot be nil"), nil)
+	}
+	for _, binding := range p.Bindings {
+		if binding.Condition != nil {
+			return nil, errwrap.Wrap(errors.New("Bigquery Datasets do not support conditional IAM"), nil)
```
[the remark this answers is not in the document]
onetwopunch: I'm not sure I'd call it "break" as much as we'd need to update this to allow conditions once they are supported, if they are ever supported (which I frankly don't think they will. be). I think it's important to let users know that attempts to apply conditions to a BQ dataset will not work as will other resources with IAM policies.

plugin/iamutil/dataset_resource.go
```diff
@@ -0,0 +1,144 @@
...
+	return policy, nil
+}
+
+func policyAsDataset(p *Policy) (*Dataset, error) {
+	ds := &Dataset{
+		Etag: p.Etag,
+	}
+
+	if p == nil {
+		return nil, errwrap.Wrap(errors.New("Policy cannot be nil"), nil)
```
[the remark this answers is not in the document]
onetwopunch: Ahh right, good point

onetwopunch:
@emilymye thanks for the review. Updated the `errwrap` calls and merged with master. Yes I was able to run the tests and also loaded the plugin into my own Vault cluster to test it.

go.mod
```diff
@@ -3,16 +3,24 @@ module github.com/hashicorp/vault-plugin-secrets-gcp
 go 1.12
 
 require (
+	github.com/Bowery/prompt v0.0.0-20190916142128-fa8279994f75 // indirect
```
kalafut, line 6: Please run `go mod tidy`.

plugin/iamutil/resource.go
```diff
@@ -0,0 +1,52 @@
+package iamutil
+
+import (
+	"context"
+	"github.com/hashicorp/go-gcp-common/gcputil"
+)
+
+// IamResource handles constructing HTTP requests for getting and
```
kalafut, line 8:
```suggestion
// Resource handles constructing HTTP requests for getting and
```

plugin/iamutil/resource.go
```diff
@@ -0,0 +1,52 @@
...
+	GetConfig() *RestResource
+	GetRelativeId() *gcputil.RelativeResourceName
+}
+
+type RestResource struct {
+	// Name is the base name of the resource
+	// i.e. for a GCE instance: "instance"
+	Name string
+
+	// Type Key is the identifying path for the resource, or
```
kalafut, line 22:
```suggestion
	// TypeKey is the identifying path for the resource, or
```

plugin/iamutil/resource.go
```diff
@@ -0,0 +1,52 @@
...
+	// Service is the name of the service this resource belongs to.
+	Service string
+
+	// IsPreferredVersion is true if this version of the API/resource is preferred.
+	IsPreferredVersion bool
+
+	// IsPreferredVersion is true if this version of the API/resource is preferred.
+	GetMethod RestMethod
+
+	// IsPreferredVersion is true if this version of the API/resource is preferred.
```
kalafut, line 37: Update comments

kalafut:
@onetwopunch Can you please prepare a docs PR in the main Vault repo linking back to this one?

plugin/iamutil/api_handle.go
```diff
@@ -0,0 +1,117 @@
...
+	relId := r.GetRelativeId()
+	replacementMap := make(map[string]string)
+
+	if strings.Contains(restMethod.Path, "{+resource}") {
+		// +resource is used to represent full relative resource name
+		if len(config.Parameters) == 1 && config.Parameters[0] == "resource" {
+			relName := ""
+			tkns := strings.Split(config.TypeKey, "/")
+			for _, colId := range tkns {
+				relName += fmt.Sprintf("%s/%s/", colId, relId.IdTuples[colId])
```
kalafut, line 101: Should we add a check that `colId` is in the map to avoid `"foo//"` and perhaps obscure errors later on?

plugin/iamutil/dataset_resource.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return policy, nil
+}
+
+func policyAsDataset(p *Policy) (*Dataset, error) {
+	ds := &Dataset{
+		Etag: p.Etag,
+	}
+
+	if p == nil {
```
kalafut, line 82: This check should be moved to the top, since currently line 79 will panic.

plugin/iamutil/dataset_resource.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+	return ds, nil
+}
+
+func datasetAsPolicy(ds *Dataset) *Policy {
+	policy := &Policy{
+		Etag: ds.Etag,
+	}
+	if ds == nil {
```
kalafut, line 117: This should be at the top of the function.

plugin/iamutil/dataset_resource_test.go
```diff
@@ -0,0 +1,240 @@
+package iamutil
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/hashicorp/go-gcp-common/gcputil"
```
kalafut, line 6: Please run `goimports`

plugin/iamutil/iam_resource_test.go
```diff
@@ -73,7 +75,18 @@ func TestParsedIamResource(t *testing.T) {
...
-	setR, err := r.SetIamPolicyRequest(expectedP)
+
+	jsonP, err := json.Marshal(expectedP)
+	if err != nil {
+		t.Fatalf("Could not json marshal expected policy: %v", err)
+	}
+
+	reqJson := fmt.Sprintf(r.config.SetMethod.RequestFormat, jsonP)
+	if !json.Valid([]byte(reqJson)) {
+		t.Fatalf("Could not formate expected policy: %v", err)
```
kalafut, line 86:
```suggestion
		t.Fatalf("Could not format expected policy: %v", err)
```

plugin/iamutil/iam_resource_test.go
```diff
@@ -227,7 +240,18 @@ func TestConditionalIamResource(t *testing.T) {
...
```
kalafut, line 251:
```suggestion
		t.Fatalf("Could not format expected policy: %v", err)
```

plugin/iamutil/dataset_resource.go
onetwopunch: Good catch!

plugin/iamutil/api_handle.go
onetwopunch: Sure happy to add this for cleanliness. However since the map it's looking at is static, I don't think this would end up with unforeseen errors. Good callout either way tho

onetwopunch:
@kalafut thanks for the review! Just updated with fixes to your review comments after running `goimports` and `go mod tidy`. I'll add an example to the docs repo tomorrow and link it here.

onetwopunch:
@kalafut Added the PR to the Vault repo: https://github.com/hashicorp/vault/pull/8728

onetwopunch:
@kalafut Any update on this?

PR #87, opened 2020-05-18
Truncate ServiceAccount display names longer than 100 characters

### LINKED ISSUE #68: Rolesets with long names cannot be updated because SA display name exceeds 100 characters
We are autogenerating roleset names from other entity data and they can be >50 characters in length.

Vault generates a description for the service account that reads `Service account for Vault secrets backend role set <name>` and adds 51 characters of boilerplate. This overflows the 100 character length limit for the description and causes service account creation to fail.

mdgreenfield:
This will sha256(displayName[92:]) and concat the first 8
characters of the sha256 hash to the display name.

Fixes #68

### DIFF
--- a/plugin/role_set.go
+++ b/plugin/role_set.go
@@ -22,8 +22,9 @@ import (
 )
 
 const (
-	serviceAccountMaxLen          = 30
-	serviceAccountDisplayNameTmpl = "Service account for Vault secrets backend role set %s"
+	serviceAccountMaxLen            = 30
+	serviceAccountDisplayNameMaxLen = 100
+	serviceAccountDisplayNameTmpl   = "Service account for Vault secrets backend role set %s"
 )
 
 type RoleSet struct {
@@ -304,7 +305,7 @@ func (rs *RoleSet) addWALsForCurrentAccount(ctx context.Context, s logical.Stora
 func (rs *RoleSet) newServiceAccount(ctx context.Context, s logical.Storage, iamAdmin *iam.Service, project string) (string, error) {
 	saEmailPrefix := roleSetServiceAccountName(rs.Name)
 	projectName := fmt.Sprintf("projects/%s", project)
-	displayName := fmt.Sprintf(serviceAccountDisplayNameTmpl, rs.Name)
+	displayName := roleSetServiceAccountDisplayName(rs.Name)
 
 	walId, err := framework.PutWAL(ctx, s, walTypeAccount, &walAccount{
 		RoleSet: rs.Name,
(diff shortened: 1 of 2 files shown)


plugin/role_set_test.go
```diff
@@ -0,0 +1,30 @@
+package gcpsecrets
+
+import (
+	"fmt"
+	"testing"
+)
+
+func Test_roleSetServiceAccountDisplayName(t *testing.T) {
```
agy, line 8: Style suggestion: I think that [table tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests) would be cleaner and clearer here.

emilymye (dismissed):

mdgreenfield:
@emilymye is this something can get merged?

plugin/role_set.go
```diff
@@ -415,6 +416,18 @@ func roleSetServiceAccountName(rsName string) (name string) {
 	return name
 }
 
+func roleSetServiceAccountDisplayName(name string) string {
+	hashLen := 8
+	fullDisplayName := fmt.Sprintf(serviceAccountDisplayNameTmpl, name)
+	displayName := fullDisplayName
+	if len(fullDisplayName) > serviceAccountDisplayNameMaxLen {
+		truncIndex := serviceAccountDisplayNameMaxLen - hashLen
+		h := fmt.Sprintf("%x", sha256.Sum256([]byte(fullDisplayName[truncIndex:])))
```
calvn, line 425: Should the sha256sum be calculated off the full `fullDisplayName` to reduce the chances of a collision or is there a particular reason we're only considering bytes within `fullDisplayName[truncIndex:]`? The final length on the display name seems to be truncated by `h[:hashLen]` anyways.

plugin/role_set.go
```diff
@@ -415,6 +416,18 @@ func roleSetServiceAccountName(rsName string) (name string) {
 	return name
 }
 
+func roleSetServiceAccountDisplayName(name string) string {
+	hashLen := 8
```
calvn, line 420: Can we move this into the const declarations up top?

plugin/role_set.go
calvn: Actually, the first part of the display name is constructed via `fullDisplayName[:truncIndex]` so even if the hash ends up being the same as another one that had the same suffix string, the first part would be unique if the full display name is different from the one that it "collided" with so I think it's fine.

calvn (dismissed):

plugin/role_set.go
mdgreenfield: fixed

calvn (approved):

PR #104, opened 2021-01-21
Fix WAL entries persisting when gcp project no longer exists

Valarissa:
# Overview
This change seeks to remove lingering WAL issues, where WAL entries would continue to persist after the GCP project
was deleted. It seeks to prevent repeated API calls to GCP attempting to delete entities that no longer exist in GCP due
to a previous cleanup. The issue would manifest for some users where they would begin to receive errors from google saying they had attempted too many API calls, and were being throttled.

Outside of fixing that behavior, there should be no change to the user experience.

This will not have an update in docs because it is restoring a behavior that was assumed to exist already.

# Design of Change
This change fixed some of the logic surrounding the WALs of iam_policy and account objects, checking, at various points, for 404 or 403 errors being returned from Google's API.

# Contributor Checklist
[x] Backwards compatible

### DIFF
--- a/plugin/rollback.go
+++ b/plugin/rollback.go
@@ -199,6 +199,9 @@ func (b *backend) serviceAccountPolicyRollback(ctx context.Context, req *logical
 
 	p, err := r.GetIamPolicy(ctx, apiHandle)
 	if err != nil {
+		if isGoogleAccountNotFoundErr(err) || isGoogleAccountUnauthorizedErr(err) {
+			return nil
+		}
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


plugin/rollback.go
```diff
@@ -297,11 +302,13 @@ func isGoogleApiErrorWithCodes(err error, validErrCodes ...int) bool {
 	if err == nil {
 		return false
 	}
-	gErr, ok := err.(*googleapi.Error)
+	ok := errwrap.ContainsType(err, new(googleapi.Error))
```
calvn, line 305: Curious on this, was there an issue with type assertion on the error?

calvn (approved):
Looks good!

plugin/rollback.go
Valarissa: So, that was necessary because the method to grab the iam_policy was wrapped with errwrap. By doing so, it always returned false. So I needed to check all of the errors to make sure they weren't googleapi errors that were wrapped, and then grab them and cast them as googleapi errors.

PR #100, opened 2020-11-05
Refactor roleset into utils mk II

lawliet89:
## Overview

Hi @kalafut, this is a clean up of @emilymye's PR in https://github.com/hashicorp/vault-plugin-secrets-gcp/pull/62

I have done the following:

- Merge `master` in to fix conflicts
- Fix some integration tests

If you could get this merged in, I will then proceed to rebase https://github.com/hashicorp/vault-plugin-secrets-gcp/pull/67 and hopefully get that in too. I have a WIP rebase [here](https://github.com/lawliet89/vault-plugin-secrets-gcp/compare/roleset_refactor...lawliet89:static_account).

## Acceptance Tests Output

```
$ make test-acc
?       github.com/hashicorp/vault-plugin-secrets-gcp   [no test files]
=== RUN   TestConfigRotateRootUpdate
=== PAUSE TestConfigRotateRootUpdate
=== RUN   TestConfig
=== PAUSE TestConfig
=== RUN   TestPathRoleSet_Basic
--- PASS: TestPathRoleSet_Basic (19.09s)
=== RUN   TestPathRoleSet_UpdateKeyRoleSet
--- PASS: TestPathRoleSet_UpdateKeyRoleSet (28.79s)
=== RUN   TestPathRoleSet_RotateKeyRoleSet
--- PASS: TestPathRoleSet_RotateKeyRoleSet (27.80s)
=== RUN   TestPathRoleSet_UpdateTokenRoleSetScopes
--- PASS: TestPathRoleSet_UpdateTokenRoleSetScopes (22.50s)
=== RUN   TestPathRoleSet_UpdateTokenRoleSet
--- PASS: TestPathRoleSet_UpdateTokenRoleSet (34.81s)
=== RUN   TestPathRoleSet_RotateTokenRoleSet
--- PASS: TestPathRoleSet_RotateTokenRoleSet (32.50s)
=== RUN   TestSecrets_GenerateAccessToken
--- PASS: TestSecrets_GenerateAccessToken (15.56s)
=== RUN   TestSecrets_GenerateKeyConfigTTL
2020/11/18 16:39:47 [DEBUG] test check failed with error Get "https://iam.googleapis.com/v1/projects/yw-testing/roles?alt=json&prettyPrint=false": oauth2: cannot fetch token: 400 Bad Request
Response: {"error":"invalid_grant","error_description":"Invalid JWT Signature."} (attempt 0), sleeping one second before trying again
[...]

PR #107, opened 2021-02-24
Add Support for Static Account

### LINKED ISSUE #60: GCP secrets: please support issuing keys/tokens for static service accounts

lawliet89:
# Overview

Fixes #60.

This is a cleanup of the original PR https://github.com/hashicorp/vault-plugin-secrets-gcp/pull/67.

It adds support for Vault to use pre-created GCP Service Accounts and issue tokens/keys.

There are some scenarios where the environment is wary of allowing automated tools like Vault access to `x.setIamPolicy` permissions. This PR allows users to create these service accounts first, and then let Vault issue credentials.

- #100 

```
$ make test-acc
?       github.com/hashicorp/vault-plugin-secrets-gcp   [no test files]
ok      github.com/hashicorp/vault-plugin-secrets-gcp/plugin    399.185s
ok      github.com/hashicorp/vault-plugin-secrets-gcp/plugin/iamutil    3.171s
?       github.com/hashicorp/vault-plugin-secrets-gcp/plugin/iamutil/internal   [no test files]
ok      github.com/hashicorp/vault-plugin-secrets-gcp/plugin/util       0.005s
?       github.com/hashicorp/vault-plugin-secrets-gcp/scripts/gohelpers [no test files]
```

Docs will be added to upstream Vault once the code is reviewed and merged in.

# Contributor Checklist
[ ] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
[x] Backwards compatible


lawliet89:
Hey @calvn and @kalafut, this is the second part of the static account PR that @emilymye started working on a while ago.

husunal:
@lawliet89 thanks for working on this. This is one of the key features that we would love to see in Vault's GCP secret backend.

lawliet89:
@husunal You can actually compile this yourself and load it up into Vault as a plugin to use it. Process is a bit cumbersome though. See https://www.vaultproject.io/docs/internals/plugins

I have a build for Linux AMD64 [here](https://github.com/lawliet89/vault-plugin-secrets-gcp/releases/tag/static-2020-11-19) but for a security product like this, I think it's better off if you can compil;e and verify it yourself.

lawliet89:
Hey @calvn and @kalafut, could I get this looked at please?

peter-fe:
@lawliet89, thx for pushing that forward. Lovely feature that would perfectly fit CI/CD needs, having a static SA in the project with all needed permissions and you could get a (short lived) token for that account via Vault in the build pipelines.
This decoupling would massively ease the current process of having those permissions reflected through temp. Vault created SAs.

Sinble:
@lawliet89 100% agree with @peter-fe . Watching closely for this merge.

kalafut:
@lawliet89 Yes, I'd like to review this one soon. Thank you for your work on it!

plugin/path_static_account_secrets.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+func pathStaticAccountSecretServiceAccountKey(b *backend) *framework.Path {
+	return &framework.Path{
+		Pattern:    fmt.Sprintf("%s/%s/key", staticAccountPathPrefix, framework.GenericNameRegex("static_account")),
+		Deprecated: true,
+		Fields: map[string]*framework.FieldSchema{
+			"static_account": {
```
austingebauer, line 16: I suggest to replace `static_account` here with `name` along with what's in `GenericNameRegex()` above. This is more consistent with what's in `pathStaticAccount`. This also applies to `pathStaticAccountSecretAccessToken`.

plugin/path_static_account_secrets.go
```diff
@@ -0,0 +1,110 @@
...
+	"fmt"
+
+	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+func pathStaticAccountSecretServiceAccountKey(b *backend) *framework.Path {
+	return &framework.Path{
+		Pattern:    fmt.Sprintf("%s/%s/key", staticAccountPathPrefix, framework.GenericNameRegex("static_account")),
+		Deprecated: true,
```
austingebauer, line 14: Is this supposed to be marked as deprecated?

plugin/path_static_account_secrets.go
```diff
@@ -0,0 +1,110 @@
...
+		},
+		HelpSynopsis:    pathServiceAccountKeySyn,
+		HelpDescription: pathServiceAccountKeyDesc,
+	}
+}
+
+func pathStaticAccountSecretAccessToken(b *backend) *framework.Path {
+	return &framework.Path{
+		Pattern:    fmt.Sprintf("%s/%s/token", staticAccountPathPrefix, framework.GenericNameRegex("static_account")),
+		Deprecated: true,
```
austingebauer, line 48: Is this supposed to be marked as deprecated?

plugin/path_static_account_secrets.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+func pathStaticAccountSecretServiceAccountKey(b *backend) *framework.Path {
+	return &framework.Path{
+		Pattern:    fmt.Sprintf("%s/%s/key", staticAccountPathPrefix, framework.GenericNameRegex("static_account")),
+		Deprecated: true,
+		Fields: map[string]*framework.FieldSchema{
+			"static_account": {
+				Type:        framework.TypeString,
+				Description: "Required. Name of the static_account.",
```
austingebauer, line 18: I suggest to stick with a common description for the field. The description is `Name to refer to this static account in Vault` above and `Name of the static_account` here. This also applies to `pathStaticAccountSecretAccessToken`.

plugin/path_static_account_secrets.go
```diff
@@ -0,0 +1,110 @@
...
+	return &framework.Path{
+		Pattern:    fmt.Sprintf("%s/%s/token", staticAccountPathPrefix, framework.GenericNameRegex("static_account")),
+		Deprecated: true,
+		Fields: map[string]*framework.FieldSchema{
+			"static_account": {
+				Type:        framework.TypeString,
+				Description: "Required. Name of the static_account.",
+			},
+		},
+		ExistenceCheck: b.pathStaticAccountExistenceCheck,
```
austingebauer, line 55: Any particular reason for this existence check? I don't think it's needed for this particular API. The existence check is often used to distinguish between create and update operations, which I don't see being done here.

plugin/path_static_account_secrets.go
```diff
@@ -0,0 +1,110 @@
...
+				Type:        framework.TypeString,
+				Description: fmt.Sprintf(`Private key type for service account key - defaults to %s"`, privateKeyTypeJson),
+				Default:     privateKeyTypeJson,
+			},
+			"ttl": {
+				Type:        framework.TypeDurationSecond,
+				Description: "Lifetime of the service account key",
+			},
+		},
+		ExistenceCheck: b.pathStaticAccountExistenceCheck,
```
austingebauer, line 35: Any particular reason for this existence check? I don't think it's needed for this particular API (similar to my other comment).

austingebauer (changes requested):
Thanks carrying this forward, @lawliet89! Just started reviewing this and left some initial comments. I'll circle back for another review shortly.

Are you able to provide some Vault CLI usage examples of the feature? I was able to get most of the way there, but I had some trouble generating a service account key for a static account. Examples would be super helpful and valuable context for the PR.

lawliet89:
Hi @austingebauer thanks for the review. I'll go through your feedback in a bit.

In the meantime, I've cooked up a demo with Terraform on how you can configure a Vault server with a custom build of this plugin installed: https://github.com/lawliet89/vault-static-demo

```
$ vault read gcp-static-test/static/test/key
Key                 Value
---                 -----
lease_id            gcp-static-test/static/test/key/OvWX6161yc3k9QDWVOIRZuyu
lease_duration      1h
lease_renewable     true
key_algorithm       KEY_ALG_RSA_2048
key_type            TYPE_GOOGLE_CREDENTIALS_FILE
private_key_data xxxxxxx
```

plugin/path_static_account.go
```diff
@@ -0,0 +1,346 @@
...
+	return &framework.Path{
+		Pattern: fmt.Sprintf("%s/%s", staticAccountPathPrefix, framework.GenericNameRegex("name")),
+		Fields: map[string]*framework.FieldSchema{
+			"name": {
+				Type:        framework.TypeString,
+				Description: "Required. Name to refer to this static account in Vault.",
+			},
+			"secret_type": {
+				Type:        framework.TypeString,
+				Description: fmt.Sprintf("Type of secret generated for this account. Defaults to %q", SecretTypeAccessToken),
```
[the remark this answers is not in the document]
lawliet89: Trying to keep it consistent with the other [non static rolesets](https://github.com/hashicorp/vault-plugin-secrets-gcp/blob/ffc5951c0d40a185352876e1a2209e353d4f4fc0/plugin/path_role_set.go#L28.

To be fair, tokens are more secure but not every API endpoint on GCP supports it (for eg signing GCS urls). I can change it if you think the default should be keys.

plugin/path_role_set.go
```diff
@@ -157,7 +157,11 @@ func (b *backend) pathRoleSetRead(ctx context.Context, req *logical.Request, d *
 }
 
 func (b *backend) pathRoleSetDelete(ctx context.Context, req *logical.Request, d *framework.FieldData) (resp *logical.Response, err error) {
-	rsName := d.Get("name").(string)
+	nameRaw, ok := d.GetOk("name")
```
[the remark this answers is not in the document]
lawliet89: I have removed this.

plugin/static_account.go
```diff
@@ -0,0 +1,343 @@
...
+	}
+
+	gcpAcct, err := b.getServiceAccount(iamAdmin, &gcputil.ServiceAccountId{
+		Project:   gcpServiceAccountInferredProject,
+		EmailOrId: input.serviceAccountEmail,
+	})
+
+	if err != nil {
+		if isGoogleAccountNotFoundErr(err) {
+			return fmt.Errorf("unable to create static account - service account %q should exist", input.serviceAccountEmail)
```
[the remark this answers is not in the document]
lawliet89: Changed this to be a comma.

plugin/path_static_account.go
```diff
@@ -0,0 +1,346 @@
...
+				Type:        framework.TypeString,
+				Description: "Email of the GCP service account to manage.",
+			},
+			"bindings": {
+				Type:        framework.TypeString,
+				Description: "Bindings configuration string.",
+			},
+			"token_scopes": {
+				Type:        framework.TypeCommaStringSlice,
+				Description: fmt.Sprintf(`List of OAuth scopes to assign to access tokens generated under this account. Ignored if secret_type not %q`, SecretTypeKey),
```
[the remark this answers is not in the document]
lawliet89: Fixed the type.

plugin/path_static_account_secrets.go
lawliet89: Removed the deprecation.

plugin/path_static_account_secrets.go
lawliet89: Removed the deprecation.

plugin/path_static_account_secrets.go
lawliet89: I have renamed this.

plugin/path_static_account_secrets.go
lawliet89: Removed the snake case.

plugin/path_static_account_secrets.go
lawliet89: I've removed it. I am not really familiar with Vault's internals.

plugin/path_static_account.go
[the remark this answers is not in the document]
austingebauer: Keeping it as `SecretTypeAccessToken` makes sense for consistency with existing rolesets 👍

austingebauer:
Thanks, @lawliet89! I was able to successfully generate access tokens and service account keys using static accounts. Happy to help you move the PR forward. I'm going to take another pass over the code for review today.

plugin/static_account.go
```diff
@@ -0,0 +1,344 @@
...
+	return walIds, nil
+}
+
+func (b *backend) addWalsForStaticAccountBindings(ctx context.Context, req *logical.Request, a *StaticAccount, removed, added ResourceBindings) (walIds []string, err error) {
+	walIds = make([]string, 0, len(removed)+len(added))
+
+	// Add WALs for resources in added bindings
+	for resource, rolesAdded := range added {
+		walEntry := &walIamPolicyStaticAccount{
+			AccountId:  a.ServiceAccountId,
```
austingebauer, line 292: I think we'll want to add `StaticAccount: a.Name` to this WAL entry. If not, there is an eventual panic in `serviceAccountPolicyDiffRollback()` (rollback.go#L263). I observed this when updating bindings on a static account to a role that caused an error.

plugin/static_account.go
```diff
@@ -0,0 +1,344 @@
...
+		walIds = append(walIds, walId)
+	}
+
+	// Add WALs for resources in removed bindings not in added bindings.
+	for resource, rolesRemoved := range removed {
+		if _, ok := added[resource]; ok {
+			continue
+		}
+		walEntry := &walIamPolicyStaticAccount{
+			AccountId:    a.ServiceAccountId,
```
austingebauer, line 312: Similar to comment above, I think we'll want to add `StaticAccount: a.Name` to this WAL entry.

plugin/rollback.go
```diff
@@ -219,59 +244,65 @@ func (b *backend) serviceAccountPolicyRollback(ctx context.Context, req *logical
...
-		}
+		return err
 	}
-	return nil
-}
-
-func (b *backend) deleteTokenGenKey(ctx context.Context, iamAdmin *iam.Service, tgen *TokenGenerator) error {
-	if tgen == nil || tgen.KeyName == "" {
-		return nil
+	if sa.ResourceName() == entry.AccountId.ResourceName() {
```
austingebauer, line 263: We should add a nil check here on `sa` to avoid a panic.

austingebauer (changes requested):

plugin/rollback.go
lawliet89: I've added a check and fixed the missing `StaticAccount` fields in the WAL entries.

austingebauer: Ty!

plugin/field_data_utils.go
```diff
@@ -0,0 +1,105 @@
...
+	if ok {
+		scopes, castOk := v.([]string)
+		if !castOk {
+			return nil, fmt.Errorf("scopes unexpected type %T, expected []string", v)
+		}
+		input.scopes = scopes
+	}
+
+	if input.secretType == SecretTypeAccessToken && len(input.scopes) == 0 {
+		return nil, fmt.Errorf("non-empty token_scopes must be provided for generating access token secrets")
```
austingebauer, line 75: nit: consistency in referring to token scopes in error message. It's `token_scopes` here and `token scopes` below.

plugin/rollback.go
```diff
@@ -219,59 +244,68 @@ func (b *backend) serviceAccountPolicyRollback(ctx context.Context, req *logical
 	return err
 }
 
-func (b *backend) deleteServiceAccount(ctx context.Context, iamAdmin *iam.Service, account gcputil.ServiceAccountId) error {
-	if account.EmailOrId == "" {
-		return nil
+func (b *backend) serviceAccountPolicyDiffRollback(ctx context.Context, req *logical.Request, data interface{}) error {
+	b.rolesetLock.Lock()
```
austingebauer, line 248: I'm wondering if this is the correct lock to use. WAL entries of type `walTypeIamPolicyDiff` are only used for static accounts, which use the backend's `staticAccountLock`.

austingebauer (changes requested):
Slowly making my way through all of the code. Thanks for the patience! Just a few more comments and suggestions.

plugin/rollback.go
lawliet89: I've changed this to `staticAccountLock`.

plugin/field_data_utils.go
lawliet89: I've changed both messages to use snake case.

plugin/field_data_utils.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	input.serviceAccountEmail = email
+	return nil, nil
+}
+
+func (input *inputParams) parseOkInputTokenScopes(d *framework.FieldData) (warnings []string, err error) {
+	// Parse secretType if not yet parsed
+	if input.secretType == "" {
+		warnings, err = input.parseOkInputTokenScopes(d)
```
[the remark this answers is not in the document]
lawliet89: Fixed. This was in the original PR that I missed.

plugin/rollback.go
```diff
@@ -219,59 +244,68 @@ func (b *backend) serviceAccountPolicyRollback(ctx context.Context, req *logical
...
+		return err
 	}
-	return nil
-}
-
-func (b *backend) deleteTokenGenKey(ctx context.Context, iamAdmin *iam.Service, tgen *TokenGenerator) error {
-	if tgen == nil || tgen.KeyName == "" {
-		return nil
+	if sa == nil {
+		return fmt.Errorf("static account %s not found", entry.StaticAccount)
```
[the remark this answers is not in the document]
lawliet89: I've made the changes. Please let me know if you would like to change the warning message.

plugin/gcp_account_resources.go
```diff
@@ -131,6 +150,146 @@ func (b *backend) createServiceAccount(ctx context.Context, req *logical.Request
...
+	}
+	return
+}
+
+func (b *backend) deleteServiceAccount(ctx context.Context, iamAdmin *iam.Service, account gcputil.ServiceAccountId) error {
+	if account.EmailOrId == "" {
+		return nil
+	}
+
+	_, err := iamAdmin.Projects.ServiceAccounts.Delete(account.ResourceName()).Do()
```
austingebauer, line 251: I realize this is existing code that's been moved to a new file, but I think we should be passing the context through to external API calls. Looks like this can be done with `.Context(ctx).Do()`.

Here are some locations (there may be more) where the context is unused:
- `deleteServiceAccount()`
- `createNewTokenGen()`
- `createServiceAccount()`
- `deleteTokenGenKey()`
- `secretKeyRevoke()`

lawliet89: Sure. I've added them.

plugin/path_role_set_secrets.go
```diff
@@ -0,0 +1,144 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+var fieldSchemaRoleSetServiceAccountKey = map[string]*framework.FieldSchema{
```
austingebauer, line 11: nit: Perhaps we can return the values set to `fieldSchemaRoleSetServiceAccountKey` and `fieldSchemaRoleSetAccessToken` from functions instead of introducing global vars? That would reduce the potential for modification at run time.

plugin/path_static_account.go
```diff
@@ -0,0 +1,346 @@
...
+	}, nil
+}
+
+func (b *backend) pathStaticAccountDelete(ctx context.Context, req *logical.Request, d *framework.FieldData) (*logical.Response, error) {
+	nameRaw, ok := d.GetOk("name")
+	if !ok {
+		return logical.ErrorResponse("name is required"), nil
+	}
+	name := nameRaw.(string)
+
```
austingebauer, line 131: I'm wondering if we should also take the `staticAccountLock` for deletion. A concurrent update to the static account could leave the service account with unexpected bindings or keys.

plugin/field_data_utils.go
```diff
@@ -0,0 +1,105 @@
...
+	rawBindings string
+	bindings    ResourceBindings
+
+	project             string
+	serviceAccountEmail string
+
+	scopes []string
+}
+
+func (input *inputParams) parseOkInputSecretType(d *framework.FieldData) (warnings []string, err error) {
```
calvn, line 25: Is there a particular reason we want the return values to be named? I also don't see `warnings` set in many of these parsing functions (aside from `parseOkInputTokenScopes`).

plugin/gcp_account_resources.go
```diff
@@ -128,7 +147,147 @@ func (b *backend) createServiceAccount(ctx context.Context, req *logical.Request
...
+
+// tryDeleteGcpAccountResources creates WALs to clean up a service account's
+// bindings, key, and account (if removeServiceAccount is true)
+func (b *backend) tryDeleteGcpAccountResources(ctx context.Context, req *logical.Request, boundResources *gcpAccountResources, removeServiceAccount bool, walIds []string) []string {
+	if boundResources == nil {
+		b.Logger().Debug("skip deletion for nil roleset resources")
+		return nil
+	}
+
+	b.Logger().Debug("try to delete GCP account resources", "bound resources", boundResources, "remove service account", removeServiceAccount)
```
calvn, line 161:
```suggestion
	b.Logger().Debug("try to delete GCP account resources", "bound_resources", boundResources, "remove_service_account", removeServiceAccount)
```

The keys on logger tend to be snake cased.

plugin/rollback.go
```diff
@@ -93,26 +101,46 @@ func (b *backend) serviceAccountKeyRollback(ctx context.Context, req *logical.Re
...
+			if sa.SecretType != SecretTypeAccessToken {
+				// Remove WAL entry - we don't clean keys if roleset generates key secrets.
+				return nil
+			}
+			if sa.TokenGen != nil {
+				keyInUse = sa.TokenGen.KeyName
+			}
 		}
+	default:
+		b.Logger().Error("removing invalid walAccountKey with empty RoleSet and empty StaticAccount - may need manual cleanup: %v", entry)
```
calvn, line 142: Nit:
```suggestion
		b.Logger().Error("removing invalid walAccountKey with empty RoleSet and empty StaticAccount, may need manual cleanup: %v", entry)
```

plugin/field_data_utils.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+}
+
+func (input *inputParams) parseOkInputServiceAccountEmail(d *framework.FieldData) (warnings []string, err error) {
+	email := d.Get("service_account_email").(string)
+	if email == "" && input.serviceAccountEmail == "" {
+		return nil, fmt.Errorf("email is required")
+	}
+	if input.serviceAccountEmail != "" && email != "" && input.serviceAccountEmail != email {
+		return nil, fmt.Errorf("cannot update email - currently %q", input.serviceAccountEmail)
```
calvn, line 49: Are we divulging too much information by returning the current value? I see this (`"... currently %q"`) in quite a few places.

plugin/field_data_utils.go
```diff
@@ -0,0 +1,105 @@
...
+		return nil, fmt.Errorf("invalid secret_type %q", secretType)
+	}
+}
+
+func (input *inputParams) parseOkInputServiceAccountEmail(d *framework.FieldData) (warnings []string, err error) {
+	email := d.Get("service_account_email").(string)
+	if email == "" && input.serviceAccountEmail == "" {
+		return nil, fmt.Errorf("email is required")
+	}
+	if input.serviceAccountEmail != "" && email != "" && input.serviceAccountEmail != email {
```
calvn, line 48: This check here is pretty essential on how this method behaves and how it should be called. Can we drop a comment here (or better yet, a  godoc on the func) on this behavior?

plugin/field_data_utils.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	rawBindings, castok := bRaw.(string)
+	if !castok {
+		return nil, fmt.Errorf("bindings are not a string")
+	}
+
+	bindings, err := util.ParseBindings(bRaw.(string))
+	if err != nil {
+		return nil, errwrap.Wrapf("unable to parse bindings: {{err}}", err)
```
calvn, line 98: We're moving away from `errwrap` and starting to use Go's builtin error wrapping function. Can we make changes on the other places where we're using `errwrap` as well?

```suggestion
		return nil, fmt.Errorf("unable to parse bindings: %w", err)
```

plugin/path_static_account.go
```diff
@@ -0,0 +1,346 @@
...
+		},
+		HelpSynopsis:    pathStaticAccountHelpSyn,
+		HelpDescription: pathStaticAccountHelpDesc,
+	}
+}
+
+func pathStaticAccountList(b *backend) *framework.Path {
+	// Paths for listing role sets
+	return &framework.Path{
+		Pattern: fmt.Sprintf("%ss?/?", staticAccountPathPrefix),
```
calvn, line 68: Is this supposed to be `%ss?/?`?

plugin/path_static_account_secrets.go
```diff
@@ -0,0 +1,106 @@
...
+	return &framework.Path{
+		Pattern: fmt.Sprintf("%s/%s/key", staticAccountPathPrefix, framework.GenericNameRegex("name")),
+		Fields: map[string]*framework.FieldSchema{
+			"name": {
+				Type:        framework.TypeString,
+				Description: "Required. Name of the static account.",
+			},
+			"key_algorithm": {
+				Type:        framework.TypeString,
+				Description: fmt.Sprintf(`Private key algorithm for service account key - defaults to %s"`, keyAlgorithmRSA2k),
```
calvn, line 21:
```suggestion
				Description: fmt.Sprintf(`Private key algorithm for service account key. Defaults to %s."`, keyAlgorithmRSA2k),
```

plugin/path_static_account_secrets.go
```diff
@@ -0,0 +1,106 @@
...
+				Default:     keyAlgorithmRSA2k,
+			},
+			"key_type": {
+				Type:        framework.TypeString,
+				Description: fmt.Sprintf(`Private key type for service account key - defaults to %s"`, privateKeyTypeJson),
```
calvn, line 26:
```suggestion
				Description: fmt.Sprintf(`Private key type for service account key. Defaults to %s."`, privateKeyTypeJson),
```

plugin/role_set.go
```diff
@@ -239,11 +238,14 @@ func (b *backend) saveRoleSetWithNewTokenKey(ctx context.Context, req *logical.R
...
 	}
-	if err := b.deleteTokenGenKey(ctx, iamAdmin, oldTokenGen); err != nil {
-		return errwrap.Wrapf("roleset update succeeded but got error while trying to delete old key - will be cleaned up later by WAL: {{err}}", err).Error(), nil
-	}
 
-	b.tryDeleteWALs(ctx, req.Storage, walId)
+	b.tryDeleteWALs(ctx, req.Storage, newWalId)
+	if oldTokenGen != nil {
+		if err := b.deleteTokenGenKey(ctx, iamAdmin, oldTokenGen); err != nil {
+			return errwrap.Wrapf("roleset update succeeded but got error while trying to delete old key - will be cleaned up later by WAL: {{err}}", err).Error(), nil
```
calvn, line 245:
```suggestion
			return fmt.Errorf("roleset update succeeded but got error while trying to delete old key - will be cleaned up later by WAL: %w", err), nil
```

plugin/rollback.go
calvn: I realized that a lot of existing errors uses `-` to break up the message so totally optional if we want to keep things consistent.

calvn:
We will have to update `go 1.12` in `go.mod` to be Go 1.13+ in order to use `%w`, but now may be the opportunity to do so.

plugin/path_static_account.go
```diff
@@ -0,0 +1,346 @@
...
+const (
+	staticAccountStoragePrefix       = "static"
+	staticAccountPathPrefix          = "static"
+	gcpServiceAccountInferredProject = "-"
+)
+
+func pathStaticAccount(b *backend) *framework.Path {
+	return &framework.Path{
+		Pattern: fmt.Sprintf("%s/%s", staticAccountPathPrefix, framework.GenericNameRegex("name")),
+		Fields: map[string]*framework.FieldSchema{
```
calvn, line 22: Might be good to mention which fields cannot be updated after creation in their `Description`.

plugin/secrets_service_account_key.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/hashicorp/go-gcp-common/gcputil"
+
```
calvn, line 9: Extra newline here.

plugin/secrets_service_account_key.go
```diff
@@ -17,6 +19,12 @@ const (
 	privateKeyTypeJson = "TYPE_GOOGLE_CREDENTIALS_FILE"
 )
 
+type secretKeyParams struct {
+	keyType, keyAlgorithm string
```
calvn, line 23: Nit:
```suggestion
	keyType      string
	keyAlgorithm string
```

plugin/path_static_account_secrets.go
```diff
@@ -0,0 +1,106 @@
...
+				Type:        framework.TypeString,
+				Description: "Required. Name of the static account.",
+			},
+		},
+		Operations: map[logical.Operation]framework.OperationHandler{
+			logical.ReadOperation:   &framework.PathOperation{Callback: b.pathStaticAccountAccessToken},
+			logical.UpdateOperation: &framework.PathOperation{Callback: b.pathStaticAccountAccessToken},
+		},
+		HelpSynopsis:    pathServiceAccountKeySyn,
+		HelpDescription: pathServiceAccountKeyDesc,
```
austingebauer, line 57:
```suggestion
		HelpSynopsis:    pathTokenHelpSyn,
		HelpDescription: pathTokenHelpDesc,
```

plugin/secrets_service_account_key.go
```diff
@@ -217,20 +194,16 @@ func (b *backend) getSecretKey(ctx context.Context, s logical.Storage, rs *RoleS
...
+const pathServiceAccountKeySyn = `Generate an service account private key secret.`
 const pathServiceAccountKeyDesc = `
-This path will generate a new service account private key for accessing GCP APIs.
-A role set, binding IAM roles to specific GCP resources, will be specified
-by name - for example, if this backend is mounted at "gcp", then "gcp/key/deploy"
-would generate service account keys for the "deploy" role set.
-
-On the backend, each roleset is associated with a service account under
-which secrets/keys are created.
+This path will generate a new service account key for accessing GCP APIs.
```
austingebauer, line 206: I think we should also update `pathTokenHelpSyn` and `pathTokenHelpDesc`. It would be good to mention static accounts (similar to what's done here) and replace the example that's using the `gcp/token/<roleset>` API, since that's being marked as deprecated.

plugin/path_static_account.go
calvn: Can we add a test case to cover this endpoint?

lawliet89:
Thanks @calvn and @austingebauer for the review. I'll look into it later on Friday. If I can't get around to it, I'll look into it next Monday.

plugin/field_data_utils.go
lawliet89: There are 103 instances of this. Should we do this in a separate PR instead?

plugin/path_role_set_secrets.go
lawliet89: Done in17c1921

plugin/path_static_account.go
lawliet89: Added in 51441cc

plugin/secrets_service_account_key.go
lawliet89: Removed.

plugin/field_data_utils.go
lawliet89: I've removed logging of current values from the error messages in this line and the others I can find.

plugin/field_data_utils.go
lawliet89: Added a godoc to explain this.

plugin/path_static_account.go
lawliet89: Added in a7e9ea70ce064a17ad5d14db0e192beaab44f825

plugin/secrets_service_account_key.go
lawliet89: Done in 533727eb13d6092015b229f1d87852c095ce0960

plugin/field_data_utils.go
lawliet89: > Is there a particular reason we want the return values to be named? 

I don't quite understand what you mean.

> I also don't see warnings set in many of these parsing functions (aside from parseOkInputTokenScopes).

Emily wrote the original code. I would keep it this way as well, so that there's consistency in how the `parseOk` functions are called.

plugin/path_static_account.go
lawliet89: I think the original intention was for this path to be `statics` in a similar vein to listing `rolesets`. I've removed the plural "s". I can add it in if you think that's more consistent.

Added test step to list static accounts in the existing tests in 1bdd252470b3b65796ced60b10d530430d44a64d

lawliet89:
I've answered/addressed the review comments. 

Also ran the integration tests and everything looks OK.

plugin/field_data_utils.go
calvn: Yeah, let's leave it out for now.

plugin/field_data_utils.go
calvn: Thanks!

plugin/path_static_account.go
calvn: The database engine has this under `static-roles`, would something similar or `static-rolesets` work here?

calvn: I see that the new regex is `%s?/?`, but should it be `%s/?`

plugin/role_set.go
calvn: Addressing this separately.

plugin/field_data_utils.go
calvn: That's fair. I was referring to `(warnings []string, err error)` being changed to `([]string, error)` or even `error` in the helper functions where no `warnings` are set or returned. These are internal helpers though so if it makes them easier to read and consistent I'm fine with leaving as-is.

calvn (dismissed):
One more comment around the static account path, but otherwise looks good. Thanks a ton for picking this up!

austingebauer (dismissed):
Looks good to me 👍  Thanks again for carrying this forward!

Did you happen to already prepare documentation for this feature? If not, I'd be happy to help with that contribution over in the Vault repo.

plugin/path_static_account.go
calvn: Since this is for managing IAM service account, I'm leaning towards prefixing this with `static-account`. I see what you mean on adding the `s` to match LIST on `rolesets` though... If we have it as `static-account` I'd be okay with having `%ss/?` as the path for LIST to keep that pattern consistent within the engine. We should enforce the plural naming if we're sticking with that though instead of making that optional.

plugin/field_data_utils.go
lawliet89: I think I'll leave it in as-is for now.

plugin/path_static_account.go
lawliet89: I've changed the prefix `static-account`. I left the regex as `%ss?/?` because the path to list rolesets also has the `s` as optional (`rolesets?/?`). What do you think about that?

lawliet89:
@austingebauer I've not prepared the documentation yet.

@calvn I've renamed the prefix to `static-account`. See  my reply at https://github.com/hashicorp/vault-plugin-secrets-gcp/pull/107#discussion_r665073688

Re-running the acceptance tests:

```
$ make test-acc
?       github.com/hashicorp/vault-plugin-secrets-gcp   [no test files]
ok      github.com/hashicorp/vault-plugin-secrets-gcp/plugin    363.110s
ok      github.com/hashicorp/vault-plugin-secrets-gcp/plugin/iamutil    5.718s
?       github.com/hashicorp/vault-plugin-secrets-gcp/plugin/iamutil/internal   [no test files]
ok      github.com/hashicorp/vault-plugin-secrets-gcp/plugin/util       (cached)
?       github.com/hashicorp/vault-plugin-secrets-gcp/scripts/gohelpers [no test files]
```

plugin/path_static_account.go
austingebauer: I'm good with `%ss?/?`, since that's consistent with listing rolesets. Also good with `static-account` as the prefix 👍

austingebauer (approved):
Thanks, @lawliet89. LGTM. I'm going to merge this now.

I'm happy to write the documentation for the feature. Also open to collaborating on it 👍

lawliet89:
@austingebauer I've opened a draft PR to work on the documentation: https://github.com/hashicorp/vault/pull/12027

Please feel free to make any edits.

austingebauer:
Thanks for doing that, @lawliet89! I'll work off of that branch with you on the docs.

peter-fe:
@calvn, @austingebauer can you give an rough estimate, when this will be released?
Seems the last release was at start of the year.

austingebauer:
@peter-fe - This is targeted to be released in Vault 1.8 which should go out around the end of July.

Issue #113, opened 2021-06-15
Creation of a RoleSet wipes conditions from the IAM Policy


trolleksii:
We are using both IAM Bindings with time expiry conditions to grant users temporary access to GCP resources and Vault GCP Secrets Engine to issue Service Account keys for our pipelines. We noticed that conditions disappear after the RoleSet creation/update, effectively making all our temporary access grants(including ones that already expired) permanent.

## How to reproduce:

1. Create a Project in GCP.
2. Grant yourself any Role with any condition.
3. Get the IAM Policy(`gcloud projects get-iam-policy <PROJECT_ID>`), the condition will be present.
4. Create a Roleset with at least one project-scope binding:
```
vault write gcp/roleset/my-key-roleset \
    project="<PROJECT_ID>" \
    secret_type="service_account_key"  \
    bindings=-<<EOF
      resource "//cloudresourcemanager.googleapis.com/projects/<PROJECT_ID>" {
        roles = ["roles/viewer"]
      }
EOF
```

5. Get the IAM Policy(`gcloud projects get-iam-policy <PROJECT_ID>`) again. The condition will be missing.

This example is for the Project Role Bindings but the same behavior can be observed with other resources.

It seems like the bug is caused by the `ChangedBindings` function in the [plugin/iamutil/iam_policy.go](./plugin/iamutil/iam_policy.go). It fails to copy the conditions to the updated IAM Policy Object.

I will submit the PR to fix this shortly.

trolleksii:
Prepared the [PR#114](https://github.com/hashicorp/vault-plugin-secrets-gcp/pull/114) to fix the bug.

davidcharbonnier:
Hello,
Is there any plans from Hashicorp team to take a look at this? Fix is already ready and just needs to be reviewed.
Thank you :)

maxcoulombe:
Fix was reviewed and merged. Thanks for looking into this!

I'll take care of the merge back into Vault so the fix in part of the upcoming 1.13 release.

PR #114, opened 2021-06-15
Preserve the IAM binding conditions during IAM Policy update 

trolleksii:
# Overview

This change fixes the bug described in [#113](https://github.com/hashicorp/vault-plugin-secrets-gcp/issues/113) when the IAM Policy were stripped of all conditions during the RoleSet creation/update/delete causing all condition-based IAM bindings to become permanent. This happened because the ChangedBindings function didn't capture the Condition of the initial IAM Policy.

[ ] [Issue #113](https://github.com/hashicorp/vault-plugin-secrets-gcp/issues/113)

# Contributor Checklist
[X] Add relevant docs to upstream Vault repository or sufficient reasoning why docs won’t be added yet
Documentation isn't required as this is a bugfix
[X] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
[X] Backwards compatible


gw0:
Conditions are also necessary to support giving access to specific GCP Cloud SQL DB instances (https://cloud.google.com/sql/docs/postgres/project-access-control#allow_users_to_connect_to_specific_instances) instead of giving project-wide access.

I was unable to use an instance self-link, full resource name, or relative resource name for a specific DB instance either with v1beta4 or v1. Although the docs suggest one could use all these approaches for all GCP resources (https://www.vaultproject.io/docs/secrets/gcp#bindings).

maxcoulombe (approved):
Looks good, thanks for taking the time to add tests as well.

maxcoulombe:
Hey @trolleksii,

I know it's been a little while but thanks for looking into this. I just tried it locally and I can confirm that the conditions were being removed, which is a pretty nasty side effect, and that your modification fixes the issue.

You can go ahead and merge this, I'll take care of the changelog when bumping the plugin version back into Vault during the release. Since it's been a long time and I'm not sure if you are still following this issue, I'll merge it myself by EoW if we don't hear any news.

Cheers and thanks again!

PR #115, opened 2021-07-06
Deprecate setting TTL/MaxTTL via /config path

sethvargo:
This centralizes the "ttl capping" into a central function and simultaneously deprecates said functionality to be removed in a future release. Administrators should use "vault secrets tune" to control TTLs instead of this configuration. The configuration remains parsed and used, but any paths that use the TTL now generate a warning message instructing users to migrate to "vault secrets tune" instead.

# Contributor Checklist
[My Docs PR Link](https://github.com/hashicorp/vault/pull/12001)
[x] ~Add output for any tests not ran in CI to the PR description (eg, acceptance tests)~ N/A
[x] Backwards compatible

### DIFF
--- a/plugin/path_config.go
+++ b/plugin/path_config.go
@@ -3,6 +3,7 @@ package gcpsecrets
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


sethvargo:
Closing due to lack of response.

Issue #99, opened 2020-11-04
Support Service Account Impersonation


lawliet89:
The `access_token` secret type should support [impersonating](https://cloud.google.com/iam/docs/impersonating-service-accounts#iam-service-accounts-grant-role-sa-gcloud) a chain of one or more service accounts. 

The impersonation can be done via:

- The service account that the secrets engine was mounted with
- Creating a new service account and making sure the operator configures the first service account in the impersonation chain to give the Vault created service account the right permission
- A pre-created service account as requested in https://github.com/hashicorp/vault-plugin-secrets-gcp/issues/83 and https://github.com/hashicorp/vault-plugin-secrets-gcp/issues/60

glebsts:
> Note: Only users with access to the Admin APIs can access the Admin SDK Directory API, therefore your service account needs to impersonate one of those users to access the Admin SDK Directory API.

https://developers.google.com/admin-sdk/directory/v1/guides/delegation#go


So, how could I get service token while impersonating certain user account? Otherwise a) can't get access to API b) audit is usually built per impersonated account (checked with JWT access to API), so it is kinda must have.

salrashid123:
i've implemented a variation of this FR here as a side feature of [Credential Access Boundary](https://cloud.google.com/iam/docs/downscoping-short-lived-credentials) with vault (not as part of true gcp secrets engine)

https://github.com/salrashid123/vault-plugin-secrets-gcp-cab#impersonated-token

usage is like this

```
vault write gcpcab/cab/config/mytotallyunrestrictedconfig  \
 project="$PROJECT_ID"  \
 target_service_account="$IMPERSONATED_SERVICE_ACCOUNT"  \
 scopes="https://www.googleapis.com/auth/cloud-platform"  \
 restricted=false \
 raw_token=true

vault policy write impersonated-policy -<<EOF
path "gcpcab/cab/certname62020" {
    capabilities = ["update", "delete"]
    allowed_parameters = {    
      "config" = ["mytotallyunrestrictedconfig"]
  }
}
EOF

# acquire impersonated token
vault token create -policy=impersonated-policy

export VAULT_TOKEN=...
vault write gcpcab/cab/certname62020 config="mytotallyunrestrictedconfig"
$ vault write gcpcab/cab/certname62020 config="mytotallyunrestrictedconfig"
Key             Value
---             -----
[...]

PR #129, opened 2021-12-02
Support impersonated accounts

mdgreenfield:
# Overview
This PR introduces a new set of endpoints (under `gcp/impersonated-account/`) allowing callers to generate a GCP access token without requiring Vault to create/store/retrieve a GCP service account key for each configured role.

This is particularly useful for GCP service accounts that, for whatever reason, have 10 keys (the GCP hard limit).

While the [newish static accounts feature](https://github.com/hashicorp/vault-plugin-secrets-gcp/issues/60) is much welcomed, it unfortunately creates a service account key every time a service account is configured as a new role/static account. Meaning that if the same service account is used with X Vault roles/static accounts then X keys are also created on that service account.

With this changeset, Vault does not authenticate with GCP as each configure service account, instead the plugin uses its own configured service account and uses [GCP account impersonation feature](https://cloud.google.com/iam/docs/impersonating-service-accounts) to generate an access token for the respective service account.

Because of this, the plugin's configured service account needs access to create access tokens for the particular service account. This is done by assigning the `roles/iam.serviceAccountTokenCreator` to the impersonated service account allowing the plugin's own service account to create the access token. This looks like the following permissions:

```
resource "google_service_account_iam_member" "allow_example_sa_to_be_impersonated" {
  service_account_id = google_service_account.example_sa.name

  member = "serviceAccount:${google_service_account.vault_gcp_secrets_backend.email}"
  role   = "roles/iam.serviceAccountTokenCreator"
}
```

# Design of Change
The static accounts code was used as a starting point. However, instead of adding a new set of endpoints the static accounts code could instead be altered to support different `secret_type`. We opt'd to create a separate set of endpoints to make maintaining our forked plugin less complicated.

- [Issue #99](https://github.com/hashicorp/vault-plugin-secrets-gcp/issues/99)
- https://github.com/salrashid123/vault-plugin-secrets-gcp#impersonated

# Contributor Checklist
- [ ] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
[My Docs PR Link](link)
[Example](https://github.com/hashicorp/vault/commit/2715f5cec982aabc7b7a6ae878c547f6f475bba6)
- [x] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
- [x] Backwards compatible


mdgreenfield:
If this PR is something the maintainers would like to pursue I'm happy to make changes, update documentation, and provide all required test output to move this along.

nrbunn:
I actually came to write this feature myself and found yours.  Is there anything I can do to help? Would love to see this in one of next releases.

Zlaticanin:
@mdgreenfield Hello! Thank you so much for your contribution. Is this something you are still interested in? 
We are planning to take a look at this and support it in Vault 1.13. In the meantime, could you please resolve the branch conflicts and we will start with the PR review?

mdgreenfield:
Yes, we're still very interested in this. We have been using this change across all our clusters in production for nearly a year.

I'll work to get the conflicts resolved in the next day or two.

mdgreenfield:
PR has been rebased and conflicts resolved.

plugin/path_impersonated_account_secrets.go
```diff
@@ -0,0 +1,87 @@
...
+		acctTtl = config.TTL
+	} else {
+		b.Logger().Debug("using impersonated account ttl for token",
+			"impersonated_account_name", acctName, "impersonated_account_ttl", acctTtl.String())
+	}
+
+	tokenSource, err := impersonate.CredentialsTokenSource(ctx, impersonate.CredentialsConfig{
+		TargetPrincipal: acct.EmailOrId,
+		Scopes:          acct.TokenScopes,
+		Lifetime:        time.Duration(acctTtl),
```
Zlaticanin, line 70:
```suggestion
		Lifetime:        acctTtl,
```

plugin/path_impersonated_account.go
```diff
@@ -0,0 +1,259 @@
...
+
+	"github.com/hashicorp/errwrap"
+	"github.com/hashicorp/vault/sdk/framework"
+	"github.com/hashicorp/vault/sdk/logical"
+)
+
+const (
+	impersonatedAccountStoragePrefix = "impersonated-account"
+	impersonatedAccountPathPrefix    = "impersonated-account"
+)
```
Zlaticanin, line 16: Trying to understand, is there a reason why we are declaring two constants with the same value? Can we just rename one to something like `impersonatedAccountPrefix` and use just that one instead?

austingebauer (commented):
Looking good! Left a few minor comments. Going to give this a test as well. Thanks, @mdgreenfield.

Zlaticanin (commented):
Thank you @mdgreenfield. A few small nits and comments

plugin/path_impersonated_account.go
mdgreenfield: I'm happy to do that but I based this off of https://github.com/hashicorp/vault-plugin-secrets-gcp/blob/3235c5de81b15c9a9db930651be7036e19aaf787/plugin/path_static_account.go#L14-L15

I suspect that it is done like that so that either the storage or the path prefix can be changed without impacting the other. I'm not sure how realistic a change like that is though and if that change did come about splitting the prefixes could be done then.

plugin/path_impersonated_account_secrets.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	config, err := getConfig(ctx, req.Storage)
+	if err != nil {
+		return nil, err
+	}
+
+	acctTtl := time.Duration(acct.Ttl) * time.Second
+	if acctTtl > config.MaxTTL {
+		b.Logger().Debug("using backend max ttl which is less than impersonated account ttl for token",
```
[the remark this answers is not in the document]
mdgreenfield: I think I agree with adding warnings with the response, however I don't think it is necessary to keep these as debug logs in that case since the debug logs are server logs and the warnings will end up in the audit logs.

plugin/path_impersonated_account_secrets.go
```diff
@@ -0,0 +1,87 @@
...
+			"impersonated_account_name", acctName, "impersonated_account_ttl", acctTtl.String(), "backend_max_ttl", config.MaxTTL.String())
+		acctTtl = config.MaxTTL
+	} else if acctTtl == 0 {
+		b.Logger().Debug("using backend default ttl since impersonated account ttl not configured for token",
+			"impersonated_account_name", acctName, "backend_default_ttl", config.TTL.String())
+		acctTtl = config.TTL
+	} else {
+		b.Logger().Debug("using impersonated account ttl for token",
+			"impersonated_account_name", acctName, "impersonated_account_ttl", acctTtl.String())
```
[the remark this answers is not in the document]
mdgreenfield: This log was only useful for a brief period when the changes were being developed but hasn't been relied on since so I agree with removing it.

plugin/impersonated_account_field_data_utils.go
```diff
@@ -0,0 +1,50 @@
+package gcpsecrets
+
+import (
+	"fmt"
+
+	"github.com/hashicorp/vault/sdk/framework"
+)
+
+type impersonatedAccountInputParams struct {
```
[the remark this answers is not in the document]
mdgreenfield: I believe so, yes. Looking at my history this looks like an artifact left over from some initial changes that were later removed.

mdgreenfield:
I've addressed the comments and unit tests are passing but I still need to run through the integration tests. I'll take care of that tomorrow.

mdgreenfield:
```
➜  vault-plugin-secrets-gcp git:(impersonated-account) ✗ make testacc
?       github.com/hashicorp/vault-plugin-secrets-gcp   [no test files]
ok      github.com/hashicorp/vault-plugin-secrets-gcp/plugin    175.243s
?       github.com/hashicorp/vault-plugin-secrets-gcp/plugin/cache      [no test files]
ok      github.com/hashicorp/vault-plugin-secrets-gcp/plugin/iamutil    (cached)
?       github.com/hashicorp/vault-plugin-secrets-gcp/plugin/iamutil/internal   [no test files]
ok      github.com/hashicorp/vault-plugin-secrets-gcp/plugin/util       (cached)
?       github.com/hashicorp/vault-plugin-secrets-gcp/scripts/gohelpers [no test files]
```

mdgreenfield:
Rebased with `main` and confirmed all integration tests are passing.

plugin/path_impersonated_account.go
Zlaticanin: That makes sense. Thank you, I missed that. Since it's following the existing pattern, it's good to leave as is. Thank you again! Great work!

austingebauer:
@mdgreenfield - This is looking good! Thanks for being so responsive here. Taking a final pass over this today.

Do you happen to have a documentation PR around this? It would be over in the main Vault repo. If not, we're happy to help on the documentation.

mdgreenfield:
I do not have a documentation PR for this. If you all are able to put one together I'm happy to review it.

austingebauer:
Sounds good! We can put a docs PR together and add you as a reviewer 👍

plugin/path_impersonated_account_secrets.go
```diff
@@ -0,0 +1,87 @@
...
+	} else if acctTtl == 0 {
+		warnings = append(warnings, fmt.Sprintf("using backend default ttl %q since impersonated account ttl not configured for token",
+			config.TTL.String()))
+		acctTtl = config.TTL
+	}
+
+	tokenSource, err := impersonate.CredentialsTokenSource(ctx, impersonate.CredentialsConfig{
+		TargetPrincipal: acct.EmailOrId,
+		Scopes:          acct.TokenScopes,
+		Lifetime:        acctTtl,
```
austingebauer, line 69: Can you confirm that these access tokens aren't different in terms of their lifecycle from [access tokens](https://cloud.google.com/docs/authentication/token-types#at-lifetime) generated via rolesets and static accounts? I mostly want to be sure that they shouldn't be leased secrets for a similar reason explained in [this section](https://developer.hashicorp.com/vault/docs/secrets/gcp#deprecation-of-access-token-leases) of our docs.

From reading the GCP [access token lifetime](https://cloud.google.com/docs/authentication/token-types#access) docs and testing this manually, my guess is they're the same. I'm not aware of a way to revoke these access tokens before their expiration, so it makes sense they're not leased in Vault. I supposed we could eventually support renewal via refresh tokens, but we don't need to add that now. I was able to confirm the tokens are revoked after their expiry and have the 1hr cap by default.

austingebauer (approved):
Had one last question. Otherwise, LGTM! Thanks again for this contribution, @mdgreenfield.

plugin/path_impersonated_account_secrets.go
mdgreenfield: > Can you confirm that these access tokens aren't different in terms of their lifecycle from [access tokens](https://cloud.google.com/docs/authentication/token-types#at-lifetime) generated via rolesets and static accounts?

Yes, that is correct. The access tokens returned from this work have the same lifecycle as those access tokens from rolesets and static accounts.

plugin/path_impersonated_account.go
```diff
@@ -0,0 +1,253 @@
...
+
+	initialInput := &ImpersonatedAccount{
+		Name:        acct.Name,
+		Ttl:         acct.Ttl,
+		TokenScopes: acct.TokenScopes,
+		ServiceAccountId: gcputil.ServiceAccountId{
+			Project:   acct.Project,
+			EmailOrId: acct.EmailOrId,
+		},
+	}
```
austingebauer, line 181: Kind of a nit, but I wonder if we could pass `acct` directly to `b.parseImpersonateInformation(acct, d)` instead of copying each of its fields into `initialInput`? Is there any value in doing this copy over to a new struct? We could delete lines 173-181 if there is no difference.

plugin/path_impersonated_account_secrets.go
austingebauer: Thanks! That makes sense.

plugin/path_impersonated_account.go
mdgreenfield: Nice! Cleans up really nicely and both unit and integration tests still pass.

Zlaticanin (approved):
Looking good! Thank you 👍

PR #136, opened 2022-02-03
Fix bug where error was not reported for invalid bindings

### LINKED ISSUE #135: Incorrect roleset bindings doesn't return an error
Hi all,

When we try to create a roleset with an invalid binding Vault doesn't return any error,

```
❯ vault -v                                                                                                             
Vault v1.9.2 (f4c6d873e2767c0d6853b5d9ffc77b0d297bfbdf)
```

Create a binding in an unavailable GCP project,
```
❯ vault write gcp/roleset/my-key-roleset6 \
    project="amplified-way-132415" \
    secret_type="access_token"  \
    token_scopes="https://www.googleapis.com/auth/cloud-platform" \
    bindings=-<<EOF
      resource "//cloudresourcemanager.googleapis.com/projects/unavailable-gcp-project" {
        roles = ["roles/viewer"]
      }
EOF

Success! Data written to: gcp/roleset/my-key-roleset6
```

```
2022-01-20T16:45:47.548Z [DEBUG] secrets.gcp.gcp_ea145832: updating roleset with new account
2022-01-20T16:45:47.548Z [DEBUG] secrets.gcp.gcp_ea145832: adding WALs for old roleset resources
2022-01-20T16:45:47.548Z [DEBUG] secrets.gcp.gcp_ea145832: skip WALs for nil roleset resources
2022-01-20T16:45:47.548Z [DEBUG] secrets.gcp.gcp_ea145832: adding WALs for new roleset resources
2022-01-20T16:45:47.548Z [DEBUG] secrets.gcp.gcp_ea145832: add WAL for service account key: account=projects/amplified-way-132415/serviceAccounts/vaultmy-key-roleset-1642715147@amplified-way-132415.iam.gserviceaccount.com keyName="\"\""
2022-01-20T16:45:47.548Z [DEBUG] secrets.gcp.gcp_ea145832: creating service account: project=amplified-way-132415 request="&{vaultmy-key-roleset-1642715147 0xc0002a7520 [] []}"
2022-01-20T16:45:48.766Z [DEBUG] secrets.gcp.gcp_ea145832: creating IAM bindings: account_email=vaultmy-key-roleset-1642715147@amplified-way-132415.iam.gserviceaccount.com bindings=map[//cloudresourcemanager.googleapis.com/projects/unavailable-gcp-project:map[roles/viewer:{}]]
2022-01-20T16:45:48.766Z [DEBUG] secrets.gcp.gcp_ea145832: setting IAM binding: resource=//cloudresourcemanager.googleapis.com/projects/unavailable-gcp-project roles=map[roles/viewer:{}]
2022-01-20T16:45:48.766Z [DEBUG] secrets.gcp.gcp_ea145832: getting IAM policy for resource name: name=//cloudresourcemanager.googleapis.com/projects/unavailable-gcp-project
2022-01-20T16:45:49.478Z [DEBUG] secrets.gcp.gcp_ea145832: creating new TokenGenerator (service account key): account=projects/amplified-way-132415/serviceAccounts/vaultmy-key-roleset-1642715147@amplified-way-132415.iam.gserviceaccount.com scopes=["https://www.googleapis.com/auth/cloud-platform"]
2022-01-20T16:45:50.543Z [DEBUG] secrets.gcp.gcp_ea145832: skip deletion for nil roleset resources
```

Same result if we try to create a binding on a resource where Vault's credentials doesn't have permission.

Expected behaviour:
Vault should return an error for unsuccessful bindings.

jasonodonnell:
Two errors were being swallowed causing Vault to report a successful operation, however, bindings were incorrect and it resulted in a service account with no bindings. Fixed these errors and added a test ensuring an error is returned.

Fixes #135.

### DIFF
--- a/plugin/path_role_set.go
+++ b/plugin/path_role_set.go
@@ -318,7 +318,7 @@ func (b *backend) pathRoleSetCreateUpdate(ctx context.Context, req *logical.Requ
 		warnings = append(warnings, updateWarns...)
 	}
 	if err != nil {
-		return logical.ErrorResponse(err.Error()), nil
+		return logical.ErrorResponse(err.Error()), err
 	} else if warnings != nil && len(warnings) > 0 {
 		return &logical.Response{Warnings: warnings}, nil
 	}
--- a/plugin/path_role_set.go
+++ b/plugin/path_role_set.go
@@ -318,7 +318,7 @@ func (b *backend) pathRoleSetCreateUpdate(ctx context.Context, req *logical.Requ
 		warnings = append(warnings, updateWarns...)
 	}
 	if err != nil {
-		return logical.ErrorResponse(err.Error()), err
+		return logical.ErrorResponse(err.Error()), nil
 	} else if warnings != nil && len(warnings) > 0 {
 		return &logical.Response{Warnings: warnings}, nil
 	}
--- a/plugin/gcp_account_resources.go
+++ b/plugin/gcp_account_resources.go
@@ -109,7 +109,7 @@ func (b *backend) createIamBindings(ctx context.Context, req *logical.Request, s
 		b.Logger().Debug("getting IAM policy for resource name", "name", resourceName)
 		p, err := resource.GetIamPolicy(ctx, apiHandle)
 		if err != nil {
-			return nil
+			return err
 		}
 
 		b.Logger().Debug("got IAM policy for resource name", "name", resourceName)
--- a/plugin/path_role_set_secrets_test.go
+++ b/plugin/path_role_set_secrets_test.go
@@ -20,6 +20,11 @@ func TestSecrets_getRoleSetKey(t *testing.T) {
 	testGetRoleSetKey(t, rsName, fmt.Sprintf("roleset/%s/key", rsName))
 }
 
+func TestSecrets_roleSetBadResource(t *testing.T) {
+	rsName := "test-bad-resource"
+	testGetRoleSetBadResource(t, rsName, fmt.Sprintf("roleset/%s", rsName))
+}
+
 // Test deprecated path still works
 func TestSecretsDeprecated_getRoleSetAccessToken(t *testing.T) {
 	rsName := "test-gentoken"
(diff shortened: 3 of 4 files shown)


plugin/path_role_set.go
```diff
@@ -318,7 +318,7 @@ func (b *backend) pathRoleSetCreateUpdate(ctx context.Context, req *logical.Requ
 		warnings = append(warnings, updateWarns...)
 	}
 	if err != nil {
-		return logical.ErrorResponse(err.Error()), nil
+		return logical.ErrorResponse(err.Error()), err
```
kalafut, line 321: How is `err` in the second parameter used here (I can't remember how the framework deals with it)? I noticed that a few lines up we're returning an error response with `nil`.

jasonodonnell: The CLI handles the logical response being an error, but in writing a test I had to inspect the response for the error, which felt weird instead of just checking of error was nil. See https://github.com/hashicorp/vault-plugin-secrets-gcp/pull/136/files#diff-6307edd822eb0e21be96d45f609fcf26677d56545380eb9392d625b6ac6b3cb7R151-R160

austingebauer: I'm not sure that we're swallowing an error in this case. Returning `logical.ErrorResponse` and a nil error is common for user errors.

Last I checked, the framework would only use the `logical.ErrorResponse` when it was returned along with a non-nil error. This ends up resulting in a 400 Bad Request. Returning the error along with a `logical.ErrorResponse` didn't make a difference from the perspective of an HTTP client. 

From the perspective of the test, you could instead:
```
resp, _ = testRoleSetCreateRaw()
if !resp.IsError() {
    t.Fatal("expected error, got none")
}
```

calvn: @austingebauer  is right -- we can either `return logical.ErrorResponse, nil` or `return nil, err`, but don't have to do both. On tests, it does make error-checking a bit more tedious since you'd have to do something like ` err != nil || (resp != nil && resp.IsError()` but in either case they it result in a non-200 response.

jasonodonnell: Done.

calvn (approved):

austingebauer (approved):

PR #144, opened 2022-08-01
Fixes duplicate static account key creation from performance secondary clusters

austingebauer:
## Overview

This PR fixes an issue that caused duplicate service account keys to be generated when creating a static account (`secret_type=access_token`) from a Vault performance secondary cluster. WAL writes in this secrets engine usually result in request forwarding (via [ErrReadOnly](https://github.com/hashicorp/vault/blob/main/sdk/logical/storage.go#L17)) on performance standby instances. However, the active instance in the performance secondary cluster is capable of writing WALs, so the request handler would partially succeed. This would result in duplicate services account keys once the request was finally forwarded due to a different storage write.

I've also added explicit request forwarding for other locations in the code that were implicitly forwarding requests on performance standbys since they could result in similar issues.

## Testing

I tested this fix using a performance secondary cluster. Two service account keys are no longer being generated when writing a static account.

### DIFF
--- a/plugin/path_role_set.go
+++ b/plugin/path_role_set.go
@@ -49,10 +49,14 @@ func pathRoleSet(b *backend) *framework.Path {
 				Callback: b.pathRoleSetRead,
 			},
 			logical.CreateOperation: &framework.PathOperation{
-				Callback: b.pathRoleSetCreateUpdate,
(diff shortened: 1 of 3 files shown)


vinay-gopalan (approved):
LGTM!

fairclothjm (approved):

PR #147, opened 2022-08-01
test: adds a simple retry loop for service account deletion

austingebauer:
This PR adds a simple retry loop in test code that asserts a service account has been deleted. I've noticed that the service account will eventually be deleted when stepping through the test code with a debugger. This creates less intermittent failures when running the acceptance tests for this plugin.


vinay-gopalan (approved):
Looks good to me!

PR #150, opened 2022-09-16
Adds simple retry loop to improve tests

austingebauer:
This PR improves the acceptance tests by adding a retry loop for checking that a service account key is deleted. The tests would occasionally fail without this because Google doesn't always report the service account key as being deleted immediately.

Additionally, this updates the changelog for the upcoming v0.14.0 tag.


robmonte (approved):
Looks good

PR #153, opened 2022-11-23
Fixes duplicate service account key for rotate root on standby or secondary

austingebauer:
## Overview

This PR fixes a bug where a duplicate service account key is created when a rotate root operation is taken on a Vault performance standby or secondary. The reason is that we partially complete the operation on the standby instance, encounter a read-only storage error, then forward the request to the active instance. At that point, the request handling creates another service account key.

This will get backported, tagged, and integrated into Vault.

## Contributor Checklist
- [ ] Add relevant docs to upstream Vault repository, or sufficient reasoning why docs won’t be added yet
  - Will add changelog. Other docs not required.
- [ ] Add output for any tests not ran in CI to the PR description (eg, acceptance tests)
- [x] Backwards compatible

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -82,6 +82,7 @@ local_environment_setup.sh
 
 # Local .terraform directories
 **/.terraform/*
+.terraform.lock.hcl
 
 # .tfstate files
 *.tfstate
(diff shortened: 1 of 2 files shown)


Zlaticanin (approved):
👍

vinay-gopalan (approved):

austingebauer:
@maxcoulombe - Indeed.. The best I could do was manually test using a performance standby running with a debugger. I verified that it hits [backend.go#L254](https://github.com/hashicorp/vault/blob/main/sdk/framework/backend.go#L254), the forwarding happens, and GCP doesn't have duplicate service account keys.

PR #160, opened 2022-12-17
Fixes panic in impersonated account access token handler

austingebauer:
## Overview

This PR fixes a panic that occurs when requesting an impersonated account access token before the secrets engine [config](https://developer.hashicorp.com/vault/api-docs/secret/gcp#write-config) is written. I checked other locations in the code where this could occur but didn't find any.

This code is unreleased in Vault, so there is no need to backport it here. 

## Related Issues/Pull Requests
-  #129

### DIFF
--- a/plugin/path_impersonated_account_secrets.go
+++ b/plugin/path_impersonated_account_secrets.go
@@ -45,25 +45,25 @@ func (b *backend) pathImpersonatedAccountAccessToken(ctx context.Context, req *l
 		return nil, err
 	}
 
-	config, err := getConfig(ctx, req.Storage)
+	cfg, err := getConfig(ctx, req.Storage)
 	if err != nil {
 		return nil, err
 	}
-	if config == nil {
-		return logical.ErrorResponse("configuration for secrets engine does not exist"), nil
+	if cfg == nil {
+		cfg = &config{}
 	}
 
 	warnings := []string{}
 	acctTtl := time.Duration(acct.Ttl) * time.Second
-	if acctTtl > config.MaxTTL {
+	if acctTtl > cfg.MaxTTL {
 		warnings = append(warnings, fmt.Sprintf("using backend max ttl %q which is less than impersonated account ttl %q for token",
-			config.MaxTTL.String(),
+			cfg.MaxTTL.String(),
 			acctTtl.String()))
-		acctTtl = config.MaxTTL
+		acctTtl = cfg.MaxTTL
 	} else if acctTtl == 0 {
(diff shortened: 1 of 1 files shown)


austingebauer:
@mdgreenfield - I wasn't able to request your review for some reason, so I'm requesting it here 😄

plugin/path_impersonated_account_secrets.go
```diff
@@ -49,6 +49,9 @@ func (b *backend) pathImpersonatedAccountAccessToken(ctx context.Context, req *l
 	if err != nil {
 		return nil, err
 	}
+	if config == nil {
+		return logical.ErrorResponse("configuration for secrets engine does not exist"), nil
```
austingebauer, line 53: We don't _need_ to return an error response here. For example, this isn't done in the code that creates service account key secrets [secrets_service_account_key.go#L160](https://github.com/hashicorp/vault-plugin-secrets-gcp/blob/main/plugin/secrets_service_account_key.go#L160). The config is only used for the TTL and MaxTTL in this context. Also, the docs state that you can leave the config unwritten to use application default credentials (see [step 2](https://developer.hashicorp.com/vault/docs/secrets/gcp#setup)).

Looking for thoughts on this!

robmonte: Given there's a valid use case for not writing an engine config, I'd say we probably shouldn't return an error or it inadvertently blocks off that entire path.

mdgreenfield: I definitely agree that if the config is not set this should not error out.

I think it probably makes the most sense to just not pass `Lifetime` with `impersonate.CredentialsTokenSource` and let GCP handle the default case.

Note that this is probably going to be the standard usage by most users.

For those that need to change the default of 1hr, a GCP [org policy constraint](https://cloud.google.com/resource-manager/docs/organization-policy/restricting-service-accounts#extend_oauth_ttl) of `constraints/iam.allowServiceAccountCredentialLifetimeExtension` must be allowed/enabled.

austingebauer: I agree, thanks! I've changed this in [bcd7690](https://github.com/hashicorp/vault-plugin-secrets-gcp/pull/160/commits/bcd769056647e44bf3412b08b34632766b4548c5) to not return an error if the config isn't set. I tested that the default TTL will be zero when populating the `Lifetime` field. This results in an access token with the default 1hr TTL.

mdgreenfield (approved):

austingebauer:
@robmonte @Zlaticanin - I can't re-request review for some reason so asking here. Thanks!

Zlaticanin (approved):
Sorry for the delay. LGTM! Thank you for doing this! 👍

PR #187, opened 2023-05-26
Prepare for v0.16.0 release

swenson:
Update Go version, dependencies, changelog.

Also moved `main.go` to `cmd/vault-plugin-secrets-gcp/main.go` so that we can use our tooling to release it as a separate binary.


swenson:
Thanks!

PR #189, opened 2023-07-20
Add missing `Query: true` metadata to API definitions

maxb:
As described in hashicorp/vault#21949 endpoints with non-path parameters
that can be made use of during ReadOperation callbacks should have those
fields marked as `Query: true` to ensure the OpenAPI spec truly reflects
how the endpoints can be called.

Endpoints concerned:
- gcp/key/{roleset}
- gcp/roleset/{roleset}/key
- gcp/static-account/{name}/key

The OpenAPI operation IDs for these operations that mention
"with-parameters" should also be changed, as they perpetuate
a misunderstanding - both read (GET) and update (POST/PUT) forms of
these APIs are **equally** capable of being used with parameters.


averche (approved):

averche:
Thanks again for the work, @maxb!

PR #192, opened 2023-09-01
Prepare changelog for 0.17.0 release

swenson:
Also made the acceptance tests slightly less flaky by adding a small sleep after creating a service account and changing the account name structure (to increase randomness and ensure that the account names are different, even after being truncated).


fairclothjm (approved):

swenson:
Thanks!

PR #211, opened 2024-04-25
Fix resource generation target

tomhjp:
`make update-resources` has been failing for a while now, so we haven't added new APIs in that time.

It seems like Google's `discovery/v1` client assumes docs links are in a place where they're not, which caused lots of 404s previously. It's pretty trivial to implement the functionality we need with a plain HTTP client anyway, so I replaced the calls to the API with plain HTTP calls and it's pretty close to working now. The remaining errors are:

```
errors while generating config: 
unable to add 'bigquery' (version 'v2'): unexpected method formats, get parameters: resource, set parameters: 
unable to add 'dns' (version 'v1beta2'): unable to get schema for setIamPolicy request, could not find policy in schema 'GoogleIamV1SetIamPolicyRequest'
unable to add 'dns' (version 'v1'): unable to get schema for setIamPolicy request, could not find policy in schema 'GoogleIamV1SetIamPolicyRequest'
unable to add 'poly' (version 'v1'), could not find doc - unexpected status code 502
unable to add 'realtimebidding' (version 'v1alpha'), could not find doc - unexpected status code 404
```

I've also committed the new resources that didn't error (by commenting out the early return for now).


yzdann:
We are really looking forward to see this out!

vinay-gopalan:
Closing in favor of #216

PR #214, opened 2024-05-21
Prepare for `v0.19.0` release

vinay-gopalan:


kpcraig (approved):

kpcraig:
what a weird set of errors

vinay-gopalan:
Adding a note here for reference that the errors in build are expected. The `make generate` target has been broken and upgrading to Go1.21 highlights that in the build with the `toolchain` directive. There is a draft PR #211 open to fix this. For now, manual acceptance tests, unit tests and builds (external and built-in) were used to confirm that the secrets engine works as expected.

PR #216, opened 2024-06-10
Fix resource generation target

vinay-gopalan:
# Overview
Builds on #211 and fixes remaining broken resources.

# Related Issues/Pull Requests
PR #211

PR #218, opened 2024-07-04
Fix a couple of bugs in #216

eh-steve:
#216 has a couple of bugs


vinay-gopalan (approved):
Thanks for the fix @eh-steve and sorry for the delay on the review! 🙏🏼 

The fix looks good, and I can merge these to run against the tests on upstream PR. Thanks again!

PR #287, opened 2026-08-07
Add missing OpenAPI metadata to the API endpoints

AnujPatnaik1:
Who the change affects or is for (stakeholders)?
The change affects mainly people who use the Vault's UI. With the added fields, they can better understand the endpoints in the gcp secrets plugin.

What is the change?
I added the missing operation summary, path descriptions, and proper typed response schema for the gcp secrets plugin.

Why is the change needed?
The change is needed because many of the plugins in vault were missing path descriptions, operation summaries, and typed response schemas. Many of the response schemas had a bare 200 Description OK response, which is not sufficient.

How does this change affect the user experience (if at all)?
When someone uses the Vault UI, they are able to understand the vault plugins better because they have a detailed description of what the endpoint does and meaningful response schema. With the response schema, they can understand what the endpoint returns.

# Design of Change
How was this change implemented?
I developed an AI skill that automatically adds the HelpSynopsis, Summary, and Response Fields. The Help Synopsis field helps with the path descriptions, and the Help Description and path name are used to come up with the Help Synopsis field. The path name, Operation Prefix, and Operation Suffix are used to come up with the operation summary. For the response schema, the AI skill is able to detect what data types are being added to the response data map, and the skill generates a response schema with those data types. For methods that return a nil, nil, a 204 response code is added.

### DIFF
--- a/plugin/path_config_rotate_root.go
+++ b/plugin/path_config_rotate_root.go
@@ -27,6 +27,18 @@ func pathConfigRotateRoot(b *backend) *framework.Path {
 		Operations: map[logical.Operation]framework.OperationHandler{
 			logical.UpdateOperation: &framework.PathOperation{
 				Callback:                    b.pathConfigRotateRootWrite,
+				Summary:                     "Rotate the root credentials for the GCP secrets engine.",
+				Responses: map[int][]framework.Response{
+					200: {{
+						Description: "OK",
+						Fields: map[string]*framework.FieldSchema{
+							"private_key_id": {
+								Type:        framework.TypeString,
+								Description: "ID of the new GCP service account key.",
+							},
+						},
+					}},
+				},
 				ForwardPerformanceStandby:   true,
 				ForwardPerformanceSecondary: true,
 			},
@@ -141,9 +153,7 @@ func (b *backend) rotateRootCredential(ctx context.Context, req *logical.Request
 	return nil
 }
 
-const pathConfigRotateRootHelpSyn = `
-Request to rotate the GCP credentials used by Vault
-`
+const pathConfigRotateRootHelpSyn = `Request to rotate the GCP credentials used by Vault.`
 
 const pathConfigRotateRootHelpDesc = `
 This path attempts to rotate the GCP service account credentials used by Vault
--- a/plugin/path_config_rotate_root.go
+++ b/plugin/path_config_rotate_root.go
@@ -26,8 +26,8 @@ func pathConfigRotateRoot(b *backend) *framework.Path {
 
 		Operations: map[logical.Operation]framework.OperationHandler{
 			logical.UpdateOperation: &framework.PathOperation{
-				Callback:                    b.pathConfigRotateRootWrite,
-				Summary:                     "Rotate the root credentials for the GCP secrets engine.",
+				Callback: b.pathConfigRotateRootWrite,
+				Summary:  "Rotate the root credentials for the GCP secrets engine.",
 				Responses: map[int][]framework.Response{
 					200: {{
 						Description: "OK",
--- a/plugin/path_impersonated_account.go
+++ b/plugin/path_impersonated_account.go
@@ -46,15 +46,51 @@ func pathImpersonatedAccount(b *backend) *framework.Path {
 		Operations: map[logical.Operation]framework.OperationHandler{
 			logical.DeleteOperation: &framework.PathOperation{
 				Callback: b.pathImpersonatedAccountDelete,
+				Summary:  "Delete an impersonated account.",
+				Responses: map[int][]framework.Response{
+					204: {{Description: "No Content"}},
+				},
 			},
 			logical.ReadOperation: &framework.PathOperation{
 				Callback: b.pathImpersonatedAccountRead,
+				Summary:  "Return an impersonated account.",
+				Responses: map[int][]framework.Response{
+					200: {{
+						Description: "OK",
+						Fields: map[string]*framework.FieldSchema{
+							"service_account_project": {
+								Type:        framework.TypeString,
+								Description: "GCP project of the impersonated service account.",
+							},
+							"service_account_email": {
+								Type:        framework.TypeString,
+								Description: "Email of the impersonated GCP service account.",
+							},
+							"token_scopes": {
+								Type:        framework.TypeSlice,
+								Description: "List of OAuth scopes assigned to access tokens.",
+							},
+							"ttl": {
+								Type:        framework.TypeInt,
+								Description: "Lifetime of the token in seconds.",
+							},
+						},
+					}},
+				},
 			},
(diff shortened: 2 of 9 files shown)


plugin/path_config_rotate_root.go
```diff
@@ -26,7 +26,19 @@ func pathConfigRotateRoot(b *backend) *framework.Path {
...
 		Operations: map[logical.Operation]framework.OperationHandler{
 			logical.UpdateOperation: &framework.PathOperation{
-				Callback:                    b.pathConfigRotateRootWrite,
+				Callback: b.pathConfigRotateRootWrite,
+				Summary:  "Rotate the root credentials for the GCP secrets engine.",
+				Responses: map[int][]framework.Response{
+					200: {{
+						Description: "OK",
+						Fields: map[string]*framework.FieldSchema{
+							"private_key_id": {
```
fairclothjm, line 35: TIL that this is returned in the rotate-root response! I checked and this is not documented on our official docs either.

plugin/path_impersonated_account.go
```diff
@@ -46,15 +46,51 @@ func pathImpersonatedAccount(b *backend) *framework.Path {
...
+								Description: "Email of the impersonated GCP service account.",
+							},
+							"token_scopes": {
+								Type:        framework.TypeSlice,
+								Description: "List of OAuth scopes assigned to access tokens.",
+							},
+							"ttl": {
+								Type:        framework.TypeInt,
+								Description: "Lifetime of the token in seconds.",
+							},
```
fairclothjm, line 76: I noticed that some of these types are different to those on line 27 for the same fields. Is that intentional?

plugin/path_impersonated_account.go
```diff
@@ -46,15 +46,51 @@ func pathImpersonatedAccount(b *backend) *framework.Path {
...
+					204: {{Description: "No Content"}},
+				},
 			},
 			logical.ReadOperation: &framework.PathOperation{
 				Callback: b.pathImpersonatedAccountRead,
+				Summary:  "Return an impersonated account.",
+				Responses: map[int][]framework.Response{
+					200: {{
+						Description: "OK",
+						Fields: map[string]*framework.FieldSchema{
```
fairclothjm, line 60: Looking at line 27, we are basically duplicating code. I wonder if we could refactor this such that if we ever add a new field, we only have to update 1 place in the code vs 2 places? Have you explored that?

plugin/path_role_set_secrets.go
```diff
@@ -59,15 +59,55 @@ func pathRoleSetSecretServiceAccountKey(b *backend) *framework.Path {
 		Operations: map[logical.Operation]framework.OperationHandler{
 			logical.ReadOperation: &framework.PathOperation{
 				Callback: b.pathRoleSetSecretKey,
+				Summary:  "Generate a service account key for a roleset.",
 				DisplayAttrs: &framework.DisplayAttributes{
 					OperationSuffix: "roleset-key2",
```
fairclothjm, line 98: Is this OperationSuffix correct? I know you didn't add this. But I am wondering if it should be something like "roleset-key-read"?

plugin/path_role_set_secrets.go
```diff
@@ -59,15 +59,55 @@ func pathRoleSetSecretServiceAccountKey(b *backend) *framework.Path {
...
 				},
+				Responses: map[int][]framework.Response{
+					200: {{
+						Description: "OK",
+						Fields: map[string]*framework.FieldSchema{
```
fairclothjm: It looks like these are duplicated for the Read and Update. Should we refactor the code for reusability?

fairclothjm (commented):
Looking great! I left a few questions on OperationSuffix and reusability/code hygiene

plugin/path_impersonated_account.go
AnujPatnaik1: Yes. The types on line 27 are the parameters for creating or updating an impersonated account. The types in lines 61-76 are what's returned when reading an impersonated account.

plugin/path_impersonated_account.go
AnujPatnaik1: I looked into it, and what we need to do is define a helper function with the response schema. However, I wouldn't do it for a situation like this where the schema is mostly similar. I would refactor if the schemas were exact duplicates.

plugin/path_role_set_secrets.go
AnujPatnaik1: Yes, we should refactor the code. I was looking at the path_role_secrets.go file, and they use helper functions like fieldSchemaRoleSetServiceAccountKey() and fieldSchemaRoleSetAccessToken(), so that the same schema doesn't need to typed multiple times. I think we should keep that consistent throughout the code.

plugin/path_role_set_secrets.go
AnujPatnaik1: From what I understand, looking at the rest of the code, the 2 at the end means it's a read operation. For example, in path_static_account_secrets.go on line 95, there is a 2 at the end of the suffix for the read operation.

fairclothjm: Yes, that might be the standard that this repo follows. But is that the standard that all plugins should follow? Is it standard for openapi in general? What are the recommended best practices? Does it make sense for it to be more human-readable?

plugin/path_impersonated_account_secrets.go
```diff
@@ -32,51 +49,25 @@ func pathImpersonatedAccountSecretAccessToken(b *backend) *framework.Path {
...
+						Fields:      responseFieldsImpersonatedAccountAccessToken(),
 					}},
 				},
 			},
 			logical.UpdateOperation: &framework.PathOperation{
 				Callback: b.pathImpersonatedAccountAccessToken,
 				Summary:  "Generate an access token for an impersonated account.",
 				DisplayAttrs: &framework.DisplayAttributes{
-					OperationSuffix: "impersonated-account-access-token2",
+					OperationSuffix: "impersonated-account-access-token",
```
fairclothjm: is this a breaking change? Should these be considered immutable or is there no effect?

plugin/path_role_set_secrets.go
AnujPatnaik1: It's not standard for openapi. I think it's best to have the actual word rather than a number. Also, looking at the auth cf, I think the numbering pattern might be something that's unique to gcp. However, it looks like changing the operation suffix is a breaking change for the UI forms because the typescript files use the operationIDs.  If they change, then the forms won't be shown properly.

plugin/path_impersonated_account_secrets.go
AnujPatnaik1: Looking at the change, I will change it back. Changing the OperationSuffix changes the OperationID and if that changes, then the UI will be affected.

fairclothjm (approved):
LGTM

PR #289, opened 2026-08-21
Fixed datatypes and removed list response schemas

AnujPatnaik1:
Fixed the TypeInt data type for the response fields.

### DIFF
--- a/plugin/path_config.go
+++ b/plugin/path_config.go
@@ -57,11 +57,11 @@ func pathConfig(b *backend) *framework.Path {
 						Description: "OK",
 						Fields: map[string]*framework.FieldSchema{
 							"ttl": {
-								Type:        framework.TypeInt,
+								Type:        framework.TypeInt64,
 								Description: "Default lease for generated keys, in seconds.",
 							},
 							"max_ttl": {
-								Type:        framework.TypeInt,
+								Type:        framework.TypeInt64,
 								Description: "Maximum lifetime of generated keys, in seconds.",
 							},
 							"service_account_email": {
@@ -73,7 +73,7 @@ func pathConfig(b *backend) *framework.Path {
 								Description: "Audience of plugin identity tokens.",
 							},
 							"identity_token_ttl": {
-								Type:        framework.TypeInt,
+								Type:        framework.TypeInt64,
 								Description: "Time-to-live of plugin identity tokens, in seconds.",
 							},
 							"rotation_schedule": {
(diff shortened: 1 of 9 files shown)


plugin/path_config.go
```diff
@@ -73,19 +73,19 @@ func pathConfig(b *backend) *framework.Path {
...
+								Type:        framework.TypeInt64,
 								Description: "Time-to-live of plugin identity tokens, in seconds.",
 							},
 							"rotation_schedule": {
 								Type:        framework.TypeString,
 								Description: "CRON-style schedule for automated root credential rotation.",
 							},
 							"rotation_window": {
-								Type:        framework.TypeInt,
+								Type:        framework.TypeFloat,
```
rana: What do you think of `TypeDurationSecond` for this one?
```suggestion
								Type:        framework.TypeDurationSecond,
```

It's a little confusing since a float is emitted, but the semantics is that of _time_. 

```go
data["rotation_window"] = role.StaticAccount.RotationWindow.Seconds()
```

The other support is that there are at least three examples of using `TypeDurationSecond` in the [Azure plugin](https://github.com/hashicorp/vault-plugin-auth-azure/blob/52dfacac23339f776f194cdd8d0a30b0e401ab7e/path_config.go#L160), [Vault SDK](https://github.com/hashicorp/vault/blob/main/sdk/helper/automatedrotationutil/fields.go#L387-L395) and [database plugin](https://github.com/hashicorp/vault/blob/main/sdk/helper/automatedrotationutil/fields.go#L387-L395) too.

plugin/path_config.go
```diff
@@ -73,19 +73,19 @@ func pathConfig(b *backend) *framework.Path {
...
 								Description: "Time window in seconds for automated rotation to complete.",
 							},
 							"rotation_period": {
-								Type:        framework.TypeInt,
+								Type:        framework.TypeFloat,
```
rana: Same idea here.

What do you think of `TypeDurationSecond` for this one?
```suggestion
								Type:        framework.TypeDurationSecond,
```

It's a little confusing since a float is emitted, but the semantics is that of _time_. 

```go
data["rotation_period"] = role.StaticAccount.RotationPeriod.Seconds()
```

The other support is that there are at least three examples of using `TypeDurationSecond` in the [Azure plugin](https://github.com/hashicorp/vault-plugin-auth-azure/blob/52dfacac23339f776f194cdd8d0a30b0e401ab7e/path_config.go#L160), [Vault SDK](https://github.com/hashicorp/vault/blob/main/sdk/helper/automatedrotationutil/fields.go#L387-L395) and [database plugin](https://github.com/hashicorp/vault/blob/main/sdk/helper/automatedrotationutil/fields.go#L387-L395) too.

rana (commented):
Hi Anuj,

Thanks for making those changes. I see you correct a bunch of the similar patterns. Just found two minor changes. Rest looks good! 🎉

rana (approved):
Thank you.

